PR #111, opened 2024-01-14
feature: Embed Block

### LINKED ISSUE #107: Social media embeds
# User story

As a content editor,
I want to embed content from social media,
so that I can easily enrich a web page with external content.

# Possible solution

Use [oembed](https://en.wikipedia.org/wiki/OEmbed), an open format designed to allow embedding content from a website into another page. [OEmbed is supported by many providers](https://oembed.com/providers.json) (also available as [npm package](https://www.npmjs.com/package/oembed-providers)). An OEmbed endpoint returns HTML that can be embedded for a given resource URL (like a URL of a tweet on Twitter).

Example OEmbed usage by [`astro-embed-twitter`](https://github.com/delucis/astro-embed/tree/main/packages/astro-embed-twitter):

```astro
---
import './Tweet.css';

export interface Props {
	id: string;
}
const { id } = Astro.props;

async function fetchTweet(id: string) {
	try {
		const oembedUrl = new URL('https://publish.twitter.com/oembed');
		oembedUrl.searchParams.set('url', id);
		oembedUrl.searchParams.set('omit_script', 'true');
		oembedUrl.searchParams.set('dnt', 'true');
		return (await fetch(oembedUrl).then((res) => res.json())) as {
			url: string;
			author_name: string;
			author_url: string;
			html: string;
		};
	} catch (e) {
		console.error(
			`[error]  astro-embed
         ${e.status} - ${e.statusText}: Failed to fetch tweet ${id}`
		);
	}
}

const tweet = await fetchTweet(id);
---

{tweet && <astro-embed-tweet set:html={tweet.html} />}
```

# References

- [Embeds in Wordpress](https://wordpress.org/documentation/article/embeds/)
- [Iframely](https://github.com/itteco/iframely)

# Questions?

- Generic embed or only for specific services like Twitter and Instagram? (each provider will need it's own bit of CSS)
- Should we create a DatoCMS plugin that fetches the OEmbed data when editors enter a URL ([prior art](https://github.com/novemberfiveco/datocms-plugin-external-audio-field/)), rather than doing this during build- or run-time? Update: 🚧 https://github.com/voorhoede/datocms-plugin-oembed

jbmoelker:
# Changes

- ...
- [ ] Allowlist providers?
- [ ] Sanatise HTML / scripts?

# Associated issue

Resolves #107 

# How to test

# Checklist


src/blocks/EmbedBlock/EmbedBlock.astro
```diff
@@ -0,0 +1,58 @@
...
+        <DefaultEmbed {data} {url} />
+      )
+    ) : (
+      <a
+        class='embed-fallback'
+        href={block.url}
+        target='_blank'
+        rel='noopener noreferrer'
+      >
+        [ unable to load from {providerName} ]<br />
```
jbmoelker, line 40: Is this fallback still needed? Maybe a basic version when no `data.html` is available? If we keep this fallback, then this line should be translated.

PR #139, opened 2024-05-08
129-log-warning-for-missing-translation

### LINKED ISSUE #129: Log warning for missing translation
If there is no translation available for a given translation key the [`t` function from `lib/i18n.ts`](https://github.com/voorhoede/head-start/blob/d8b6aba09ebe06c4c19b8f978efd2cfed7d3c894/src/lib/i18n.ts#L14) will return an empty string `''`. Some translations may be used for visually hidden content like `aria-labels`, where it will be extra difficult to detect missing translations. Therefore it would be a good enhancement to log a warning for missing translations.

luukbrauckmann:
# Changes

- Adds console warning to the translation function when there is no translation for that key

# Associated issue

Resolves #129 

# How to test

1. Open project on localhost
2. Add a random translation somewhere in the home page
3. There should be a yellow console warning in your terminal (Not in the browser console)

# Checklist

- [x] ~~I have added a decision log entry if the change affects the architecture or changes a significant technology~~ 

### DIFF
--- a/src/lib/i18n.ts
+++ b/src/lib/i18n.ts
@@ -11,7 +11,12 @@ const i18n = rosetta(messages);
 i18n.locale(defaultLocale);
 
 export type T = typeof i18n.t & ((key: TranslationKey) => string);
-export const t: T = i18n.t.bind(i18n);
+export const t: T = (key: TranslationKey) => {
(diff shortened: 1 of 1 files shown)


decrek (approved):
Nice addition!

PR #143, opened 2024-06-07
fix(VideoEmbedBlock): correct layer order (z-index)

jbmoelker:
# Changes

Correct layer order (z-index) of elements in Video Embed Block.

# Associated issue

N/A

# How to test

I experienced this issue while developing the No Dodos website a while back. I kept this fix as a note, but forgot how to reproduce the issue 🤷 .

# Checklist


decrek (commented):
Nothing changed in the z index order, so it is unclear to me what this PR tries to fix. I would say that it only adds left and top of 0 to the absolute positioning right?

PR #152, opened 2024-06-11
fix(Image): make sure images do not render larger than their intrinsic size

### LINKED ISSUE #151: Inline placed images should be sized according to their intrinsic size
Currently, images that are placed in structured text are sized to be as big as the container allows for. I would expect images that are smaller to be limited to their intrinsic size when uploaded.

decrek:
# Changes

- Set a max-width of 100% in order to not render images larger then their intrinsic size
- Add an imgix parameter of fit: max in order to not upscale images larger then their intrinsic size

# Associated issue

Resolves #151

# How to test

1. Open preview link and see homepage. The voorhoede logo is an a png coming from imgix. The cake is an svg with a width an height set to 100px.


jurgenbelien (approved):
:chef's kiss:

PR #153, opened 2024-06-11
Data fetching error from dato not shown in development

### LINKED ISSUE #149: Data fetching error from Dato not shown in development
When a request fails, the `data` and `error` are destructured from the `response` object, but the logging function tries to stringify that `response` object again, instead of stringifying the error. 

See https://github.com/voorhoede/head-start/blob/0de92d6a48232f7d1ddc4e7c3c9c0e326b4944de/src/lib/datocms.ts#L48

### LINKED ISSUE #150: Preview throws 500 due to Astro bug in `server` mode.
On homepage, there should be no value for `path` to route-match on. Unfortunately for some reason, the value `undefined` is used and matched on `src/pages/[locale]/[...path]/`.

This seems to be an issue with Astro: https://github.com/withastro/astro/issues/9103

decrek:
# Changes

- Add error handling for both graphql errors as fetch errors when requesting datocms

# Associated issue

Resolves #149

# How to test

1. Locally screw up a graphql file and see relevant errors
2. Locally screw up datocmsRequest in datocms.ts, e.g. not passing headers and see a relevant error instead of page data being undefined

### DIFF
--- a/src/lib/datocms.ts
+++ b/src/lib/datocms.ts
@@ -43,6 +43,10 @@ export const datocmsRequest = async <T>({ query, variables = {}, retryCount = 1 
     if (retryCount >= retryLimit) throw Error('DatoCMS request failed. Too many retries.');
     return datocmsRequest({ query, variables, retryCount: retryCount + 1 });
   }
+  
+  if (!response.ok) {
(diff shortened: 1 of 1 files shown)


jurgenbelien (approved):
Works as expected 👍

PR #156, opened 2024-06-14
Fix/142 xlinkhref is deprecated

### LINKED ISSUE #142: `xlink:href` is deprecated
https://github.com/voorhoede/head-start/blob/0de92d6a48232f7d1ddc4e7c3c9c0e326b4944de/src/components/Icon/Icon.astro#L20

`xlink:href` is a deprecated feature according to MDN Web docs. See link:

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

Should just use `href` instead or according to the documentation if you want support for really old browser you can also keep `xlink:href` next to `href`, but we shouldn't rely on it for modern browsers.

luukbrauckmann:
# Changes

- Adds `href` to the use element. Kept `xlink:href` as fallback.
- Also adds some suggestions for improving accessibility.

# Associated issue

Resolves #142 

# How to test

1. Run dev server
2. Check share link
3. Icon works 🎉 

# Checklist


luukbrauckmann:
Added some accessibility improvements. I think this would be a better way than the previous version. If someone disagrees then I'm curious why.

Issue #103, opened 2023-12-26
Test Setup


jbmoelker:
# User story

As a developer,
I want a test setup and test coverage,
so I can author code without worrying about regression.

# Checklist

- [x] Test `src/lib/` functions with [Vitest](https://docs.astro.build/en/guides/testing/#vitest); use [filesystem & request mocking](https://vitest.dev/guide/mocking) where needed.
- [x] Test `src/blocks/` and `src/components/` (where relevant) with [Astro container + Vitest](https://docs.astro.build/en/guides/testing/#vitest-and-container-api)
- [ ] Test `src/pages/` with [Playwright](https://docs.astro.build/en/guides/testing/#playwright) (focus on critical features only) and [Playwright Axe](https://playwright.dev/docs/accessibility-testing) for accessibility testing.
- [ ] Add scripts to `package.json` with a `test:unit` and `test:e2e`.
- [ ] Add option to add test to `npm run create` scripts (see [plopfile](https://github.com/voorhoede/head-start/blob/main/config/plop/plopfile.mjs))
- [x] Run in GitHub Actions step (like the linting steps).
- [x] Document test setup

jbmoelker:
We've experienced some issues with routing in recent Astro releases. So maybe it's good to have some of our tests covering this.

And probably good practice to start adding tests to our `/api/` routes.

Individual blocks, components and partials may be easier to test once [Astro container API](https://github.com/withastro/roadmap/issues/533) is available.

jbmoelker:
The [Astro container API is now generally available](https://docs.astro.build/en/reference/container-reference/).

jbmoelker:
Possible setup to test components. Focus on how to, not what we should and shouldn't test.

`src/blocks/Image.test.ts`:
```ts
import { expect, test } from 'vitest';
import { renderToFragment, renderToString } from '@lib/renderer';
import Image from './Image.astro';

// test using rendered component as a string:
test('Image ...', async () => {
  const result = await renderToString(Image, {
    props: {
      image: {
        url: 'https://example.com/image.jpg',
        alt: 'An image',
        // title: 'A title',
      }
    },
  });

  expect(result).toContain(' alt="An image"');
  expect(result).toContain(' loading="lazy"');
  expect(result).not.toContain('<figcaption ');
  expect(result).not.toContain('</figcaption>');
});

// test using rendered component as a document fragment with DOM methods:
test('Image fragment', async () => {
  const fragment = await renderToFragment(Image, {
    props: {
[...]

PR #169, opened 2024-09-09
feat/unit-testing

WesselSmit:
# Changes

- adds `test` and `unit-test` scripts
- adds unit tests for `i18n` lib function
- adds a `vitest.config.ts` so we can configure the testing suite
- adds new deps needed for testing
- adds `unit-tests` step to github action CI so we test each commit to the repo from now on
  - uses the `github actions reporter` to enable formatting and highlighting of testing output in the CI pipeline
- adds documentation for testing setup
- fixes the following issues in the i18n lib function:
  - interpolation did not work in the `t` function
  - params and language props were not supported in `t` function

# Associated issue

Part of #103 

# How to test

- verify that the tests in the CI were succesful

# Checklist


.github/workflows/node-ci.yml
```diff
@@ -86,8 +119,44 @@ jobs:
...
+          name: dist
+          path: dist
 
       - name: Validate HTML
         run: npm run lint:html
+
+  unit-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: [setup, build]
```
jbmoelker, line 134: Why do tests need a full build? Is it because you need the files generated in `src/lib/`? Would `npm run prep` on `pretest` or `pretest:unit` suffice?

.github/workflows/node-ci.yml
```diff
@@ -27,6 +27,39 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci --ignore-scripts --no-audit --no-fund
 
+  build:
```
jbmoelker, line 30: Need job setup 😍 . Not sure if we need it for tests. See comment on that job :)

vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+/// <reference types="vitest" />
+import { getViteConfig } from 'astro/config';
+
+export default getViteConfig({
+  test: {
+    // Vitest configuration options
```
jbmoelker, line 6: Would be nice to add the [Github Actions reporter](https://vitest.dev/guide/reporters#github-actions-reporter) so our code is annotated like this:

Code annotated with test feedback

vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+/// <reference types="vitest" />
+import { getViteConfig } from 'astro/config';
```
jbmoelker, line 2: This causes the process to hang after execution:

```
 Test Files  1 passed (1)
      Tests  6 passed (6)
   Duration  499ms (transform 51ms, setup 0ms, collect 62ms, tests 7ms, environment 0ms, prepare 133ms)

close timed out after 1000ms
Tests closed successfully but something prevents Vite server from exiting
You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters
```

Hanging reporter doesn't help much:
```
There are 12 handle(s) keeping the process running

# Tinypool
file:///.../head-start/node_modules/tinypool/dist/esm/index.js:37              
file:///.../head-start/node_modules/tinypool/dist/esm/index.js:58              
file:///.../head-start/node_modules/tinypool/dist/esm/index.js:952             
file:///.../head-start/node_modules/vitest/dist/vendor/cli-api.E07AF1Yq.js:8192
file:///.../head-start/node_modules/vitest/dist/vendor/cli-api.E07AF1Yq.js:8845
file:///.../head-start/node_modules/vitest/dist/vendor/cli-api.E07AF1Yq.js:8870

# FILEHANDLE
(unknown stack trace)

# FILEHANDLE
(unknown stack trace)

...
```

Changing `{ getViteConfig } from 'astro/config'` to `{ defineConfig } from 'vitest/config'` solves the hanging issue. But not sure what new problems it creates.

We could for now also limit the `teardownTimeout`:

```ts
export default defineConfig({
  test: {
    teardownTimeout: 1000,
    // Vitest configuration options
  },
});
```

WesselSmit: For now I will use `defineConfig` as we are still in the early stages and there seem to be no additional problems so far. If we run into problems down the road, we can look into the other options, but for now this feels like a 'safe enough' stage to still make changes like this.

.github/workflows/node-ci.yml
WesselSmit: This also works, so I changed it to this!

jbmoelker (approved):

PR #174, opened 2024-09-12
chore: anchor slot trim tests

jurgenbelien:
# Changes

- Adds tests for trimming of whitespace in anchor tags.

# How to test

1. See that there are new tests, and that they do not fail

# Checklist


config/plop/templates/block/Block.astro.hbs
```diff
@@ -1,7 +1,7 @@
 ---
 import type { {{ pascalCase name }}Fragment } from '@lib/types/datocms';
 
-interface Props {
+export interface Props {
```
jbmoelker, line 4: Astro does not support named exports in components. So this shouldn't work. Does it? The exported interfaces are not imported anywhere else right? So I wouldn't do this. If it works, it is undocumented. See https://docs.astro.build/en/basics/astro-components/#component-props

PR #177, opened 2024-09-16
chore: component testing

luukbrauckmann:
# Changes

- Adds unit tests for all components and blocks

# Associated issue

None

# How to test

1. Open project in editor
2. Run in terminal `npm ci`
3. Run `npm run test`
4. All tests should run.

# Checklist


src/blocks/PagePartialBlock/PagePartialBlock.test.ts
```diff
@@ -0,0 +1,236 @@
...
+            }
+          ]
+        }
+      }
+    });
+
+    expect(fragment.querySelector('details[open]')).toBeTruthy();
+  });
+
+  // Todo: Tabs are rendered on the client and therefore not testable for now.
```
luukbrauckmann, line 189: Do we still want to test this one?

vitest.config.ts
luukbrauckmann, line 1: This is something we need to look at, because the Vitest variant can't resolve TS path aliases and also can't import `.astro` files.

src/components/Breadcrumbs/Breadcrumbs.astro
```diff
@@ -18,12 +18,12 @@ const schema = {
     {
       '@type': 'ListItem',
       name: t('breadcrumbs_home'),
-      item: new URL(`/${ locale }/`, Astro.site),
+      item: new URL(`/${ locale }/`, Astro.site || 'http://localhost:5173/'),
```
jurgenbelien, line 21: What is happening here? I see that Astro.site is undefined in Vitest. I can also verify that using `vi.stubGlobal` does not resolve this. Then I would expect the `renderFragment `in the `renderer.ts` to be able to use the Astro config, which is a valid property for the `AstroContainer.create()` function. But it doesn't seem to work atm unfortunately.

I would opt to remove the fallback here, because it can have unintended consequences, and skip the breadcrumbs component for now, with a TODO stating the issue

src/components/LocaleSelector/LocaleSelector.astro
```diff
@@ -14,8 +14,8 @@ const { pageUrls } = Astro.props;
 const getPageHref = (locale: SiteLocale) => {
   const pageUrl = pageUrls.find((pageUrl) => pageUrl.locale === locale);
   return pageUrl
-    ? new URL(pageUrl.pathname, Astro.site)
-    : new URL(`/${ locale }/`, Astro.site);
+    ? new URL(pageUrl.pathname, Astro.site || 'http://localhost:5173/')
+    : new URL(`/${ locale }/`, Astro.site || 'http://localhost:5173/');
```
jurgenbelien, line 18: Same as breadcrumbs

vitest.config.ts
jurgenbelien: the vitest variant being `vitest/config`? What would be the problem with using the astro one?

jurgenbelien (changes requested):
Good job, I can imagine this took quite some effort!

I do have some remarks regarding how to structure tests and how to handle the data going into the tests. I also noticed some changes in the component to sidestep some issues with the renderer. I think it might be wise to skip those components instead of adding behavior that might have unintended consequences. 

And a final tip that I did not mention in the comments: take a look at your commit history and see if you can clean that up a bit. You have three subsequent commits called 'Refactor'. Are you refactoring things you did in this branch earlier? If so, try to squash those with the changes your made earlier, they are no longer relevant. If not, try to elaborate a bit on what you are refactoring and why. Make commits atomic and only about the issue at hand.

vitest.config.ts
luukbrauckmann: I don't know. This was something @WesselSmit implemented. What reason was there to choose the `vitest/config`?

src/blocks/ImageBlock/ImageBlock.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    expect(fragment.querySelector('figure')).toBeTruthy();
+    expect(fragment.querySelector('figcaption')).toBeTruthy();
+    expect(fragment.querySelector('figcaption')?.textContent).toBe('A test image');
+    expect(fragment.querySelector('img')).toBeTruthy();
+    expect(fragment.querySelector('img')?.getAttribute('loading')).toBe('lazy');
+    expect(fragment.querySelector('img')?.src).toBe('https://example.com/');
+    expect(fragment.querySelector('img')?.alt).toBe('A test image');
+    expect(fragment.querySelector('img')?.style.aspectRatio).toBe('150/150');
```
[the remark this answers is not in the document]
luukbrauckmann: Good point!

vitest.config.ts
jbmoelker: See https://github.com/voorhoede/head-start/pull/169#discussion_r1750738633

src/blocks/ImageBlock/ImageBlock.test.ts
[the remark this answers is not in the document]
luukbrauckmann: I have consulted the mighty ChatGPT and it mentioned some valid points. I've broken down the tests into smaller chunks and grouped logical properties together. ChatGPT mentioned that you don't want to make the tests too specific, because it will be more difficult to maintain. Testing in isolation also leads to redundant tests and you would be wasting performance because of that.

I think what I have now seems like a good way of testing

src/middleware.ts
luukbrauckmann, line 1: For some reason eslint was not having it. I don't know why? Nothing has changed here including the package lock, so it was all still running the same version of eslint. I've checked for some global stuff (Which I thought to be highly unlikely, because I've never installed eslint globally on my system and the checks also failed in the actions) but there is nothing of eslint besides de project folder on my system.

src/lib/datocms.ts
luukbrauckmann, line 1: Same here about eslint

src/lib/datocms.ts
```diff
@@ -20,7 +20,7 @@ type DatocmsRequest = {
  */
 export const datocmsRequest = async <T>({ query, variables = {}, retryCount = 1 }: DatocmsRequest): Promise<T> => {
   const headers = new Headers({
-    Authorization: getSecret('DATOCMS_READONLY_API_TOKEN'),
+    Authorization: getSecret('DATOCMS_READONLY_API_TOKEN')!,
```
luukbrauckmann, line 23: eslint thinks this value could be undefined

src/blocks/ImageBlock/ImageBlock.test.ts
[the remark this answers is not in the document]
jurgenbelien: I think this is a good balance indeed. If pressed I would say that the loading attribute is still better suited as a separate test, but you don't have to change it.

src/components/LocaleSelector/LocaleSelector.astro
```diff
@@ -11,6 +11,8 @@ interface Props {
 }
 const { pageUrls } = Astro.props;
 
+console.log(Astro.site);
+
```
jurgenbelien, line 15: 👀

vitest.config.ts
```diff
@@ -1,8 +1,18 @@
...
+  }
 });
+
+// import { defineConfig } from 'vitest/config';
+
+// export default defineConfig({
+//   test: {
+//     reporters: process.env.GITHUB_ACTIONS ? ['dot', 'github-actions'] : ['dot'],
+//   },
+// });
```
jurgenbelien, line 18: This can be removed I guess?

jurgenbelien (approved):
Looks good! I think it's time to merge it and move any enhancements to the tests to the issues.

Issue #164, opened 2024-08-22
"Sync" current dato project with dato migrations


Siilwyn:
The Dato instance is out of sync with the current migrations, we could create a new empty environment or project but it will miss the demo content.

We either manually copy over content or write a script to fill in data so this could overlap with #27.

PR #181, opened 2024-09-18
feature: Details Block

### LINKED ISSUE #171: Details block
As a content editor,
I would like to be able to add detail lists,
so I can structure things like product details.

# Checklist

- [ ] Use [Voorhoede JSON table (list) plugin](https://www.datocms.com/marketplace/plugins/i/datocms-plugin-json-table) to easily manage detail lists in the CMS
- [ ] Use `<dl>` to semantically structure the list. We could support multiple `<dd>`'s per `<dt>` if we agree on an item split character (`\n` or `;` or ...) in the CMS
- [ ] Provide a "grid" option for the block layout ([CSS grid works well for `<dl>`'s](https://piccalil.li/blog/dl-grid/), [even with multiple `<dd>`s per `<dt>`](https://gist.github.com/LeaVerou/12b1d1e078510900c7ab206da1a0ff2e?permalink_comment_id=4537764#gistcomment-4537764))

luukbrauckmann:
# Changes

- Adds new Details Block

# Associated issue

Resolves #171 

# How to test

1. Open preview link
2. Navigate to /en/demos/page-partial-layouts/
3. You can also check Dato.

# Checklist


luukbrauckmann:
I'm a bit conflicted by the name `DetailsBlock`. If I would see this I would expect to render a `details` element, but maybe that's because I'm a developer who knows html. I suggest I rename it to `DefinitionsBlock` or `DefinitionListBlock`? 

Also I added for now the | (pipe) character for separating multiple values. I could make this an option in which you can enter the split character in Dato, so the content editors can decide on a per block basis what they want to the split character to be. Does this sound like a good idea or overkill?

src/blocks/Blocks.astro
```diff
@@ -1,5 +1,6 @@
 ---
 import type { AnyBlock } from './Blocks';
+import DetailsBlock from './DetailsBlock/DetailsBlock.astro';
```
jbmoelker, line 3: About the name of the component. I get your feedback on how it doesn't match with the HTML element. But I think that's an implementation detail. Clients also call things you click on buttons while in HTML those may be anchors. To me it's important that the name of the component reflects its purpose. The aim of this component is to structure things like product details. Hence the Details Block.

src/blocks/DetailsBlock/DetailsBlock.astro
```diff
@@ -0,0 +1,45 @@
+---
+import type { DetailsBlockFragment } from '@lib/types/datocms';
+
+interface Props {
+  block: DetailsBlockFragment
+}
+
+const { block } = Astro.props;
+const splitChar = '|';
```
jbmoelker, line 9: avoid abbreviations. Rather `splitCharacter` or `separator` (as that's how the argument in `string.split` is called)

src/blocks/DetailsBlock/DetailsBlock.astro
```diff
@@ -0,0 +1,45 @@
...
+const hasMultipleValues = (value: string) => value.includes(splitChar);
+const splitValues = (value: string) => value.split(splitChar);
+---
+
+<dl class:list={['details-block', `details-block--${block.display}`]}>
+  {
+    Object.keys(block.details).map((key) => (
```
jbmoelker, line 16: I think you can simplify this to:

```js
Object.entries(block.details).map(([key, value]) =>
```
and later use
```js
value.split(separator).map(text => <dd>{ text }</dd>)
```

so you don't need helper functions and don't need to distinguish between single and multiple value. Splitting a single value will just return an array with a single item.

src/blocks/DetailsBlock/DetailsBlock.astro
```diff
@@ -0,0 +1,45 @@
...
+  block: DetailsBlockFragment
+}
+
+const { block } = Astro.props;
+const splitChar = '|';
+const hasMultipleValues = (value: string) => value.includes(splitChar);
+const splitValues = (value: string) => value.split(splitChar);
+---
+
+<dl class:list={['details-block', `details-block--${block.display}`]}>
```
jbmoelker, line 14: Styles in Astro are scoped. So you don't need to introduce classes here. We want to avoid adding a CSS strategy and leave that choice to the developers using Head Start. Have a look at how this is done in other components.

jbmoelker, line 14: Please use `block.layout`. See other components and the docs about naming: https://github.com/voorhoede/head-start/blob/main/docs/cms-content-modelling.md

src/blocks/DetailsBlock/DetailsBlock.astro
```diff
@@ -0,0 +1,45 @@
...
+  {
+    Object.keys(block.details).map((key) => (
+      <>
+        <dt>{key}</dt>
+        {hasMultipleValues(block.details[key]) ? (
+          <div>
```
jbmoelker, line 20: Is this valid HTML? I didn't know that. I still think you might be able to do without the `<div>`. See https://gist.github.com/LeaVerou/12b1d1e078510900c7ab206da1a0ff2e?permalink_comment_id=4537764#gistcomment-4537764

src/blocks/DetailsBlock/DetailsBlock.fragment.graphql
```diff
@@ -0,0 +1,4 @@
+fragment DetailsBlock on DetailsBlockRecord {
```
jbmoelker, line 1: Please make sure to also include the migration for this block in the PR. We might be blocked by #164 by this 🤷 .

jbmoelker (changes requested):
Nice work @luukbrauckmann! Left small comments.

src/blocks/DetailsBlock/DetailsBlock.astro
luukbrauckmann: Good one!

src/blocks/DetailsBlock/DetailsBlock.astro
luukbrauckmann: You're right. I thought it was valid, but I've looked it up now and found out it is not valid. I can fix it without a div.

src/blocks/DetailsBlock/DetailsBlock.astro
luukbrauckmann: Though `char` was pretty clear, because of functions like `charAt()`, but yeah I can name it `splitCharacter`. I thought that would be more clear since `separator` is a more broader term.

src/blocks/DetailsBlock/DetailsBlock.astro
luukbrauckmann: Yeah I know. I did it that way because I thought we were consciously not styling on attributes.

src/blocks/DetailsBlock/DetailsBlock.astro
```diff
@@ -0,0 +1,34 @@
...
+const splitCharacter = '|';
+---
+
+<dl data-layout={block.layout}>
+  {
+    Object.entries(block.details).map(([key, value]) => (
+      <>
+        <dt>{key}</dt>
+        {(value as string).split(splitCharacter).map((splitValue) => (
+          <dd>{splitValue}</dd>
```
jbmoelker, line 18: I would also trim the value, so in the CMS editors can write "a | b | c" instead of "a|b|c", so `splitValue.trim()`

jbmoelker:
@luukbrauckmann you've already made some good iterations on this PR 🙌. As @decrek remarked and as we discussed, we could achieve this functionality as a variation on the Page Partial Block, rather than introducing a new one. We prefer that approach as it reduces the number of components we need to maintain and footprint we provide to developers using Head Start. What prompted the discussion and is another benefit to reusing the Page Partial Block, is that it allows for richer content in the detail list items (like text formatting and links).

jbmoelker:
Stale for 3 months. Incomplete implementation: migration script is missing, UI layout can be simplified and overflow isn't handled yet. So closing this PR for now.

PR #185, opened 2024-09-24
feature: Remove background image when image is loaded

### LINKED ISSUE #132: Remove background-image when image is loaded
When loading in a png the scaled down background-image placeholder is visible through the main image. This needs to be removed when the main image is loaded.

See possible solution:
- https://github.com/voorhoede/fellow-digitals-website/blob/23-create-badges-parade/src/blocks/ImageBlock/Image.client.ts
- https://github.com/voorhoede/fellow-digitals-website/blob/db02f6239548ab036622d0ee1c2cc626279bc2a8/src/blocks/ImageBlock/Image.astro#L31

luukbrauckmann:
# Changes

- Adds a custom element with logic to remove the background image if image loads

# Associated issue

Resolves #132 

# How to test

1. Open preview link
2. Check in the inspector if images which are wrapped in an `enhanced-image` tag if they have a background image in the styles.
3. You can also go to your network tab and enable throttling to better see it happening (But with throttling the styles do not always seem to match with the real DOM, so you have te click the element away or try to disable the background image and you'll see it change and or cannot disable the rule).
 
# Checklist

### DIFF
--- a/src/blocks/ImageBlock/Image.astro
+++ b/src/blocks/ImageBlock/Image.astro
@@ -27,39 +27,42 @@ const { image, loading = 'lazy' } = Astro.props;
 const altText = image.alt ? image.alt : ''; // default to empty string for accessibility
 const imageUnavailableMessage = t('image_unavailable');
 ---
+<enhanced-image>
+  <figure>
+    {
+      image.responsiveImage
+        ? // bitmap images have Imgix generated props (responsiveImage.*):
+        <img
+          alt={ altText }
+          loading={ loading }
+          srcset={ getImageSrcSet(image.url) }
+          src={ getImageSrc(image.url) }
+          style={{
+            aspectRatio: image.responsiveImage.aspectRatio,
+            backgroundImage: `url(${image.responsiveImage.base64})`
+          }}
+          data-unavailable={ imageUnavailableMessage }
+        />
+        : // vector images do not have Imgix generated props, so using as is:
+        <img
+          alt={ altText }
+          loading={ loading }
+          src={ image.url }
+          style={{
+            aspectRatio: `${image.width}/${image.height}`,
+          }}
+          data-unavailable={ imageUnavailableMessage }
+        />
+    }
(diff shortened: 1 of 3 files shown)


src/blocks/ImageBlock/Image.astro
```diff
@@ -27,39 +27,42 @@ const { image, loading = 'lazy' } = Astro.props;
 const altText = image.alt ? image.alt : ''; // default to empty string for accessibility
 const imageUnavailableMessage = t('image_unavailable');
 ---
+<enhanced-image>
```
jbmoelker, line 30: Since the `Image` is part of the `ImageBlock` component, I would name the custom element `image-block`. That's consistent with other custom elements in the repository and predictable when someone encounters the custom element in the code that's shipped (like in the browser). I understand it's a bit different than with other blocks as it's not the top level element in `ImageBlock.astro`, but I'm okay with that :)

src/blocks/ImageBlock/Image.client.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * A custom HTML element that enhances an `<img>` element with additional behavior.
+ * Automatically removes the inline `background-image` style once the image is fully loaded.
```
jbmoelker, line 3: I would remove all other comments except for these top ones. I understand you saw the `Tabs` component as a reference. That one was based on an article on web.dev that already included all the comments. Since this script is much more straightforward, I'd remove the rest of the comments.
The only thing I would add, is the explanation why. So:

```ts
/**
 * A custom HTML element that enhances an `<img>` element with additional behaviour to remove the inline `background-image` style once the image is fully loaded. This prevents blurred placeholder backgrounds from still being visible once an image with transparency (like png or webp) is loaded.
```

jbmoelker (changes requested):
Works as desired. Only a few requests for code style improvements.

jbmoelker (approved):

PR #190, opened 2024-10-02
Fix: Layout shifts caused by images

luukbrauckmann:
# Changes

- Fixes layout shifts by setting width and height on images

# Associated issue

Issue seen and fixed in Happy Labs.

# How to test

1. Open preview link in English
2. Open inspector
3. Go to network tab
4. Enable `ignore cache` and set network throttling on
5. You'll see the image size being reserved.

# Checklist

### DIFF
--- a/src/blocks/ImageBlock/Image.astro
+++ b/src/blocks/ImageBlock/Image.astro
@@ -37,6 +37,8 @@ const imageUnavailableMessage = t('image_unavailable');
           loading={ loading }
           srcset={ getImageSrcSet(image.url) }
           src={ getImageSrc(image.url) }
+          height={image.height}
+          width={image.width}
           style={{
             aspectRatio: image.responsiveImage.aspectRatio,
             backgroundImage: `url(${image.responsiveImage.base64})`
@@ -48,6 +50,8 @@ const imageUnavailableMessage = t('image_unavailable');
           alt={ altText }
           loading={ loading }
           src={ image.url }
+          height={image.height}
+          width={image.width}
           style={{
             aspectRatio: `${image.width}/${image.height}`,
           }}
@@ -67,6 +71,7 @@ const imageUnavailableMessage = t('image_unavailable');
 img {
   display: block;
   max-width: 100%;
+  height: auto;
   background-size: cover;
 }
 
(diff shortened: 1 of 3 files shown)


jbmoelker (approved):
I've looked at different articles to better understand why this issue occurs. Either because an `img` is a "replaced element", or because the display is incorrect or because the dimensions are over constrained or because the browser can't make out the intrinsic dimensions because it contains a `srcset`, but that shouldn't matter with a specified aspect ratio. I couldn't find anything that explains the cause of our issue. I do see that your change solves it. So I'm inclined to approve and accept it :) 

Resources I checked:
- https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio
- https://jakearchibald.com/2022/img-aspect-ratio/
- https://css-tricks.com/almanac/properties/a/aspect-ratio/

@jurgenbelien do you agree?

src/blocks/ImageBlock/Image.astro
```diff
@@ -37,6 +37,8 @@ const imageUnavailableMessage = t('image_unavailable');
           loading={ loading }
           srcset={ getImageSrcSet(image.url) }
           src={ getImageSrc(image.url) }
+          height={image.height}
```
decrek, line 40: This solves the same issue as the aspect ratio inline styling on line 43. So one or the other should be enough. I did notice that the inline calculation is actually calculated instead of dumped as a string, so thats probably why there is reflow with aspect ratio in place.

decrek (changes requested):

src/blocks/ImageBlock/Image.astro
decrek: Ok just to be clear, on page load the value of the aspect ratio is not correct, but when javascript kicks in it will be set correctly, which is too late ofcourse. No idea why JS has influence on this but on page https://head-start.pages.dev/en/demos/text-image-block/ the first image first has an aspect ratio of `aspect-ratio:1.5658959537572255;` and after JS kicks in it has a value of `aspect-ratio: 1.5659 / 1;`. You can test it with JS disabled. 🤷‍♂️

decrek: And I dont know if this is actually a problem since 1.5659 / 1 is kind of the same as 1.5659

decrek: Soooo, the problem seems to be the absence of width 100% on the image in combination with `aspect-ratio`. The aspect ratio works if you add width 100% to the image.

jurgenbelien (approved):
Approved, we discussed this and leave the `width` and `height` in but remove the `aspectRatio`.

PR #194, opened 2024-10-28
test: DatoCMS lib helpers

WesselSmit:
# Changes

- adds unit tests for datocms + search lib functions 
- changes the lib folder structure and namings + refactors all imports of the lib functions 

# Associated issue

- n/a

# How to test

- run `npm run test:unit` -- also happens automatically in the github actions

# Checklist


src/lib/datocms/index.test.ts
```diff
@@ -0,0 +1,528 @@
...
+    });
+
+    test('should retry if it hits a rate-limit error', async () => {
+      const mockData = [
+        {
+          id: 1,
+          title: 'Hello World',
+          body: 'It\'s a beautiful day today.',
+        },
+      ];
```
jurgenbelien, line 125: I would define this outside of the describe so it can be reused across all tests as `singleRecordCollection` for instance

src/lib/datocms/index.test.ts
jurgenbelien, line 1: Some ideas to make this file more manageable:
* Split this up in tests for each of the main functions:
  * `./tests/datocmsCollection.test.ts`
  * `./tests/datocmsRequest.test.ts`
  * `./tests/datocmsSearch.test.ts`
  * `./tests/formatSearchResults.test.ts`
* Separate the mock server setups from the actual tests in `/tests/mockServer.ts` and export them ready to go.

jurgenbelien (changes requested):
Whew, that's a lot! I think it is wise to break this up in some smaller chunks so it gets the attention from reviewers that it deserves.

src/lib/i18n/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    test('should log a warning if translation for given key does not exist', () => {
+      // we use .mockImplementation(() => {}) to prevent the console from actually logging the warning
+      const consoleSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+
+      t('unsupported_translation_key');
+
+      expect(consoleSpy).toHaveBeenCalled();
+      expect(consoleSpy).toHaveBeenCalledWith('\x1b[33mMissing translation for key: unsupported_translation_key');
+    });
```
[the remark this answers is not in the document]
WesselSmit: Fully agree

src/lib/datocms/index.test.ts
```diff
@@ -0,0 +1,528 @@
...
+        graphql.query('MockQuery', () => {
+          return HttpResponse.json({ data: mockData });
+        }),
+      );
+
+      const response = await datocmsRequest({
+        query: parse('query MockQuery { id }')
+      });
+      expect(response).toEqual([]);
+    });
```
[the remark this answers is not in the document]
WesselSmit: I understand where this is coming from, but do not agree. Yes its less code and yes it makes it less menacing to open the test file, but at the same time it introduces logic which makes the tests more error prone and harder to read. 

I prefer it to be more explicit, you prefer it be more concise. I think there are good arguments on both sides, but for now I do not see the point in changing this code.

.graphqlrc.ts
```diff
@@ -27,7 +27,7 @@ module.exports = {
           plugins: [
             'typescript',
             'typescript-operations',
-            'typed-document-node',
+            '@graphql-codegen/typescript-document-nodes',
```
WesselSmit, line 30: I was getting errors when using the linting script. The `TypedDocumentNode` import needs to be a 'type import' instead of a 'normal import' according to the linter. This seems to be caused by the `verbatimModuleSyntax` tsconfig option which was recently added.
 
See https://github.com/voorhoede/head-start/actions/runs/11686385220/job/32542186025 for an example.

This fixes that error.

.graphqlrc.ts
```diff
@@ -44,7 +44,7 @@ module.exports = {
               FloatType: 'number',
               IntType: 'number',
               ItemId: 'string',
-              JsonField: 'unkown',
+              JsonField: 'unknown',
```
WesselSmit, line 47: This seems to be a typo which has been in the codebase for the last 15 months. Correcting this caused the EmbedBlock type to change (since it has a json field) so I also needed to fix that.

See changes in `src/blocks/EmbedBlock/EmbedBlock.astro`

jurgenbelien (approved):
Happy to have this discussion regarding verbosity. Thanks for splitting it up, it does make reviewing it easier.

src/lib/datocms/tests/datocmsSearch.test.ts
```diff
@@ -0,0 +1,169 @@
...
+}));
+
+const mockedSearchResults = [
+  {
+    type: 'search_result',
+    id: '1',
+    attributes: {
+      title: 'Test Title',
+      body_excerpt: 'Test body',
+      url: 'https://test.com',
```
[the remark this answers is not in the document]
WesselSmit: Fully agree!

PR #191, opened 2024-10-17
docs: Upgrading

jbmoelker:
# Changes

Adds documentation on how to upgrade to a newer version of Head Start by applying changes made to your own project.

# Associated issue

Resolves #86

# How to test

1. Go to an existing project based on Head Start
2. Follow the upgrade guide
3. Verify the changes are applied

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have made sure that my PR is easy to review (not too big, includes comments)
- [x] I have made updated relevant documentation files (in project README, docs/, etc)
- [x] I have added a decision log entry if the change affects the architecture or changes a significant technology
- [x] I have notified a reviewer

PR #147, opened 2024-06-09
feature: file model, download links & file redirects

jbmoelker:
# todo

- [ ] get custom file URL redirects working, like https://feat-file-model-and-link.head-start.pages.dev/my-files/example.pdf
- [ ] document in docs/routing.md
- [ ] cleanup / refactor earlier attempts

# Changes

# Associated issue

Resolves #148 
Resolves #155

# How to test

1. Open preview link
2. Navigate to the [Files demo page](https://feat-file-model-and-link.head-start.pages.dev/en/demos/file-demo/)
3. ...

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made sure that my PR is easy to review (not too big, includes comments)
- [ ] I have made updated relevant documentation files (in project README, docs/, etc)
- [ ] I have added a decision log entry if the change affects the architecture or changes a significant technology
- [ ] I have notified a reviewer

PR #195, opened 2024-10-31
feature: directly link content records (routing)

### LINKED ISSUE #160: Easier routing / record linking
See for reference: https://github.com/No-Dodos/nododos-website/commit/ea66b8d1b17ed89741c349cb03c97660b8ecceef

jbmoelker:
# Changes

Directly directly link content records instead of using a generic InternalLink component for better content editing experience. See new decision log entry for detailed motivation.

# Associated issue

Resolves #160 

# How to test

1. Open preview link
2. Tap on the link for documentation
4. Verify that this links to a valid oage
5. Verify in Dato (environment 160-link-records) that this link is a Record Reference

# Checklist


config/datocms/migrations/1730405111_linkRecords.ts
```diff
@@ -0,0 +1,150 @@
...
+      structured_text_links: {
+        on_publish_with_unpublished_references_strategy: 'fail',
+        on_reference_unpublish_strategy: 'delete_references',
+        on_reference_delete_strategy: 'delete_references',
+        item_types: ['LjXdkuCdQxCFT4hv8_ayew', 'X_tZn3TxQY28ltSyjZUGHQ'],
+      },
+    },
+  });
+
+  console.log(
```
jbmoelker, line 28: Also used this occasion to move the `slug` field back to the top. Is was dangling all the way at the bottom of the page.

config/datocms/migrations/1730405111_linkRecords.ts
```diff
@@ -0,0 +1,150 @@
...
+  await client.itemTypes.update('V80liDVtRC-UYgd3Sm-dXg', {
+    name: '\uD83E\uDDE9 Page Partial Block',
+  });
+
+  console.log('Update block model "\uD83D\uDCCA Table Block" (`table_block`)');
+  await client.itemTypes.update('0SxYNS2CR1it_5LHYWuEQg', {
+    name: '\uD83D\uDCCA Table Block',
+  });
+
+  console.log('Manage menu items');
```
jbmoelker, line 101: Also used this occasion to organise the menu and hide the schema migrations for editors.

src/blocks/TextBlock/nodes/ItemLink.astro
```diff
@@ -10,13 +17,23 @@ type Node = {
...
   node: Node;
-  link: InternalLinkFragment;
+  link: AnyPage;
 };
 
-const { node, link } = Astro.props;
-const openInNewTab = Boolean(node.meta?.find(entry => entry.id === 'target' && entry.value === '_blank'));
+const { link, node, ...props } = Astro.props;
+const locale = getLocale() as SiteLocale;
+const href = getHref({ locale, record: link });
```
jbmoelker, line 25: Is this a logical place for the `href`, `html` and `link.title` code or should we make a `RecordLink` component somewhere which is then also used by `InlineItem.astro`?

src/blocks/TextBlock/nodes/ItemLink.test.ts
```diff
@@ -1,7 +1,7 @@
 import { describe, expect, test } from 'vitest';
```
jbmoelker, line 1: This test needs to be updated to match the new related component. As suggested in a comment on ItemLink.astro that component should maybe be split up first.

src/lib/routing/page.ts
```diff
@@ -1,36 +1,43 @@
-import type { PageRecord, SiteLocale } from '@lib/types/datocms';
+import type { HomeRouteFragment, PageRouteFragment, SiteLocale } from '@lib/types/datocms';
+
+export type PageRouteForPath =
+  & Pick<PageRouteFragment, '_allSlugLocales' | 'parentPage'>
+  & { parentPage?: PageRouteForPath | null };
+
+export type PageOrHomeRoute =
```
jbmoelker, line 7: @jurgenbelien to me `routing/page.ts` is only about the `Page` model and its routing. If the routing for the `Home` model would've been more complicated, there would've been a `routing/home.ts` as well. Similarly there's a [`routing/file.ts` in #147](https://github.com/voorhoede/head-start/pull/147/files#diff-8fcebe9a1639fde4961c47b9b8e882d8e6a3f39de0744d938e5c6ee86895dfb5). In `routing/index.ts` they all come together. So in that file I would introduce this union type (and also because it's only used there, reducing the mental model). I would consider renaming it `AnyRoute` or `RecordRoute` as with for example the addition of the `File` model it would become `PageOrHomeOrFileRoute` etc.

src/lib/routing/index.ts
```diff
@@ -0,0 +1,27 @@
+import type { SiteLocale } from '@lib/types/datocms';
+import { getLocale } from '@lib/i18n';
+
+import { getPagePath, type PageOrHomeRoute } from './page';
```
jbmoelker, line 4: Up until now we've only use `import type { ... }` (instead of `import { type ... }` in this project. I thought that was needed to make the distinction between types and runtime code, but I see both are a solution to that. Do we want to stick to a specific pattern, or just mix and match? As in, probably write imports the way to produces the least amount of `import` statements :)

src/blocks/TextBlock/nodes/ItemLink.astro
jbmoelker: @jurgenbelien what do you think about this concept:
- We move this logic to a reusable component in `components/` as it's not a modular block, and the related fragments are in `lib/routing/`.
- We start with a generic `LinkToRecord` component, which we can later add more specific components next to like `LinkToFile`. By the `LinkTo` prefix they automatically stay next to each other: `components/LinkToRecord/LinkToRecord.astro` and `components/LinkToFile/LinkToFile.astro`.
- Should we bundle them in a shared component directory like `components/LinkTo/LinkToRecord.astro` and `LinkToFile.astro` or does this become confusing? Especially since a component like `LinkToRecord/` has its own `README.md`.
- I imagine their APIs becoming `<LinkToRecord record={ ... } ... >` and `<LinkToFile file={ ... } ... >`. Does that make sense to you?

src/pages/[locale]/[...path]/index.astro
```diff
@@ -1,26 +1,19 @@
...
 
 export async function getStaticPaths() {
-  interface PagesCollectionItem {
-    _allSlugLocales: Array<{ locale: SiteLocale; value: string }>;
-    parentPage: ParentPageFragment;
-  }
-  const pages = await datocmsCollection<PagesCollectionItem>({
+  const pages = await datocmsCollection<PageRouteForPath>({
     collection: 'Pages',
     fragment: `
```
Comment (jbmoelker) [inline src/pages/[locale]/[...path]/index.astro:18]: @jurgenbelien should we attempt to `print(PageRouteFragment)` (using `graphql/language/printer`) here?

CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Changelog
```
jbmoelker, line 1: @jurgenbelien the files from #191 are already in `main`. Do you know why they are in here again?

src/pages/[locale]/[...path]/index.astro
```diff
@@ -55,7 +48,7 @@ type Params = {
 const { locale, path } = Astro.params as Params;
 const variables = { locale, slug: getPageSlugFromPath(path) };
 const { page } = (await datocmsRequest<PageQuery>({ query, variables })) as {
-  page: PageRecord;
+  page: NonNullable<PageQuery['page']>; // Only NonNullable when statically generated. Handle as a 404 when this is a server route!
```
Comment (jbmoelker) [inline src/pages/[locale]/[...path]/index.astro:51]: I think we can now solve this a different way. Since v4.13 [Astro supports rewrites, including a `404` pattern](https://docs.astro.build/en/guides/routing/#rewrites). So we should be able to do this instead:

```ts
const { page } = await datocmsRequest<PageQuery>({ query, variables })

if (!page) {
  return Astro.rewrite("/404")
}
```
I haven't used this before and haven't tested this yet either :)

src/lib/routing/index.ts
jurgenbelien: I would say, when you're only importing types, add it before the braces, when importing type and code, place it inside the braces. But I can live with separating type imports from code imports altogether. But then the question is, first all type imports, then all code, or (keep it) sorted by path and alternate between type and code?

Again my (not very deep-seated) preference is:
* sort/group by import path (dependencies first, aliased source files second, relative imports last)
* inside one `import {...} from ''` import, code first (alphabetically), types second (alphabetically)
* no enforcement of this with tooling, if the order is wrong no biggie.

src/pages/[locale]/[...path]/index.astro
Comment (jurgenbelien) [reply src/pages/[locale]/[...path]/index.astro:18]: not familiar with that, but yes I was indeed wondering if such a thing existed. However we would be overfetching if we would use PageRouteFragment

CHANGELOG.md
jurgenbelien: I'll take a look, perhaps a over-eager rebase

src/pages/[locale]/[...path]/index.astro
Comment (jurgenbelien) [reply src/pages/[locale]/[...path]/index.astro:51]: Yes this is indeed what I would prefer too, and what I mentioned on Slack. But I would argue that the better implementation would be to have a function do this, either by wrapping the datocmsRequest, or reimplementing the datocmsRequest, or extending the datocmsRequest. I would prefer to park this for another PR. Here I just wanted to make sure that the type is correct, and does not throw for the routing functions.

src/blocks/TextBlock/nodes/ItemLink.astro
```diff
@@ -1,7 +1,14 @@
 ---
-import InternalLink from '@blocks/InternalLink/InternalLink.astro';
-import type { InternalLinkFragment } from '@lib/types/datocms';
+import type { SiteLocale } from '@lib/types/datocms';
+import { getLocale } from '@lib/i18n';
+import { getHref } from '@lib/routing';
 
+type AnyPage = {
```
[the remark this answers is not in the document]
jurgenbelien: It's in `@lib/routing/index.ts` and based on the grapql fragment

src/blocks/TextBlock/nodes/ItemLink.astro
jurgenbelien: Right, yeah that makes sense

jurgenbelien: Regarding the nesting, I'd prefer to keep it as flat as possible. We have around 16 components in the `@components/` folder. I feel that is still a manageable number of items.

jurgenbelien: I've kept the ItemLink component for now. I feel that having the folder `@blocks/TextBlock/nodes` map directly to the nodes in the `Text.astro` components declaration adds clarity.

jurgenbelien:
Had to switch Dato instances to show that the pipeline succeeds for last commit https://github.com/voorhoede/head-start/pull/195/commits/0d118f6706b0e51e225273f95920ad662a16f252. Switched back after that to unblock other PRs

src/components/LinkToRecord/LinkToRecord.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  test('renders text content without trailing whitespace ', async () => {
+    const fragment = await renderToFragment<LinkToRecordProps>(LinkToRecord, {
+      props,
+      slots: {
+        default: 'spacing\n',
+      },
+    });
+    expect(fragment.querySelector('a')?.textContent).toBe('spacing');
+  });
```
WesselSmit, line 32: Is this something we want to test in this file? Trimming is done in the `Link` component and if we change this implementation of the Link component, we would not want this to fail (I think). Seems to me that this is a implementation detail that we test in `Link.test.ts`.

WesselSmit (changes requested):
A lot of good stuff, but I have some small comments/questions, mainly about how we organise the code. Besides that, nice work ✨

config/datocms/migrations/1730405111_linkRecords.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  console.log(
+    'Update Slug field "Slug" (`slug`) in model "\uD83D\uDCD1 Page" (`page`)'
+  );
+  await client.fields.update('SFfpl7xiQaGe5-jy6tlPfg', { position: 2 });
+
+  console.log(
+    'Update Structured text field "Text" (`text`) in block model "\uD83D\uDCDD Text Block" (`text_block`)'
+  );
+  await client.fields.update('NtVXfZ6gTL2sKNffNeUf5Q', {
```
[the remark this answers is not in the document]
jurgenbelien: Honestly, I'm not familiar enough with the migrations (yet) to have an opinion on this at the moment.

package.json
```diff
@@ -59,7 +59,7 @@
     "promise-all-props": "^3.0.0",
     "regexparam": "^3.0.0",
     "rosetta": "^1.1.0",
-    "typescript": "^5.2.2"
+    "typescript": "^5.6.3"
```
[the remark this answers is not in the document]
jurgenbelien: See the commit message for this change to see why ;-)

src/components/LinkToRecord/LinkToRecord.test.ts
jurgenbelien: You're right, but since I've spent some time diving through the whole stack of components to find that issue, I felt it wouldn't hurt to guard against this a bit more vociferously

src/components/Link/Link.astro
```diff
@@ -0,0 +1,20 @@
...
+// Strip trailing spaces from slot contents so underlined links
+// do not extend further than the text.
+let html;
+if (Astro.slots.has('default')) {
+  html = await Astro.slots.render('default');
+  html = html.trim();
+}
+---
+
+<a {href} {target} {...props}><Fragment set:html={html} /></a>
```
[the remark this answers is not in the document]
jurgenbelien: I'll add it!

WesselSmit (approved):

PR #202, opened 2024-11-05
feature: Table of Contents

jbmoelker:
# Changes

- Adds unique html id slugs to heading elements.
- Extracts these headings and turns them into a table of contents with links to the headings.

Result:

image

# Associated issue

N/A

# How to test

# Checklist


package.json
```diff
@@ -50,7 +50,10 @@
     "datocms-listen": "^0.1.15",
     "datocms-structured-text-utils": "^2.0.4",
     "get-video-id": "^3.6.5",
+    "github-slugger": "^2.0.0",
     "globby": "^13.2.2",
+    "hast-util-from-html": "^2.0.3",
```
jbmoelker, line 55: Some of these new dependencies are interfering with the build. Will have to do some triage first because I don't yet know what the exact problem is.

src/components/HtmlIdSlugProvider/index.ts
```diff
@@ -0,0 +1,18 @@
+import GithubSlugger from 'github-slugger';
```
jbmoelker, line 1: Placing this script and export the `resetHtmlIdSlugger` in the `astro` file caused the compiler not to be able to find `slugger`. Extracting it to this file solves that issue. But the new problem is that now this script isn't really scoped to the provider anymore.

May need to use a different mechanism, like https://docs.astro.build/en/recipes/sharing-state/ ?

src/lib/routing/html-id-slugger.ts
```diff
@@ -0,0 +1,18 @@
+import GithubSlugger from 'github-slugger';
```
jbmoelker, line 1: This file should be either a lib here or in the provider component, but not both :)

src/components/HtmlIdSlugProvider/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { fromHtml } from 'hast-util-from-html';
+import { toString } from 'hast-util-to-string';
+
+const slugger = new GithubSlugger();
+
+export const resetHtmlIdSlugger = () => {
+  slugger.reset();
+};
+
+export const htmlIdSlug = ({ id, html }: { id?: string; html: string }) => {
```
jbmoelker, line 11: should add test coverage for these methods

src/components/HtmlIdSlugProvider/index.ts
jbmoelker: Or maybe we could use an actual `context` with this library: https://withastro-utils.github.io/docs/guides/context/

jbmoelker:
Implementation is blocked as the `AstroContainer`'s `renderToString` method relies on `node:path`, which is not available in the Cloudflare runtime:

```
[ERROR] [vite] x Build failed in 1.25s
[commonjs--resolver] [plugin vite:resolve] Cannot bundle Node.js built-in "node:path" imported from "node_modules/astro/dist/core/config/schema.js". Consider disabling ssr.noExternal or remove the built-in dependency.
```

Help wanted: can we render the blocks using a different method? Or should we abandon this feature?

jbmoelker:
Will be resolved by #349

PR #203, opened 2024-11-06
test: SEO lib helpers

WesselSmit:
# Changes

- modified `siteName` and `titleSuffix` seo lib getters to be functions so they always use the locale from the moment they functions are invoked (instead of using the locale from the moment the file is first executed)
- add unit tests for seo lib function

# Associated issue

- n/a

# How to test

- run `npm run test:unit` -- this also happens automatically so you could also just check if the associated pipeline succeeds

# Checklist

### DIFF
--- a/src/lib/seo/index.ts
+++ b/src/lib/seo/index.ts
@@ -8,10 +8,19 @@ export type PageUrl = {
   pathname: string,
 };
 
-const locale = getLocale();
+export const siteName = () => {
+  const locale = getLocale();
+  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
 
-export const siteName = globalSeo[locale as keyof typeof globalSeo].siteName;
-export const titleSuffix = globalSeo[locale as keyof typeof globalSeo].titleSuffix;
+  return localeSeo.siteName;
+};
+
+export const titleSuffix = () => {
+  const locale = getLocale();
+  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
+
+  return localeSeo.titleSuffix;
+};
 
 export const noIndexTag: Tag = {
   attributes: { name: 'robots' },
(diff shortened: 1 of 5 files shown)


src/lib/seo/index.ts
```diff
@@ -8,18 +8,27 @@ export type PageUrl = {
...
-export const titleSuffix = globalSeo[locale as keyof typeof globalSeo].titleSuffix;
+  return localeSeo.siteName;
+};
+
+export const titleSuffix = () => {
+  const locale = getLocale();
+  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
+
+  return localeSeo.titleSuffix;
+};
```
WesselSmit, line 23: In order to properly unit test this functionality, I would need to test in both the `en` and `nl` locales. So I would need to mock the `getLocale` function.

But mocking the `getLocale` function twice (`en` and `nl`) would result in the mocks being immediately executed after each other due to mock calls being hoisted. This happens before even one of the tests is executed.

So the happy flow for the tests would be:
1. getLocale is being mocked to return `en`
2. test that checks for `en` locale runs
3. getLocale is being mocked to return `nl`
4. test that checks for `nl` locale runs

But this is what actually happens:
1. getLocale is being mocked to return `en`
2. getLocale is being mocked to return `nl` _-- this overwrites the previous mock_
3. test that checks for `en` locale runs _-- locale should be `en`, but uses `nl`_
4. test that checks for `nl` locale runs _-- locale correctly uses `nl`_

This all happens due to the lib/seo/index file immediately invoking the `getLocale()` call when the file is run/imported. To fix this we can refactor all seo lib functions that depend on a locale to compute the locale at the time of calling the function. 

So
```ts
const locale = getLocale(); // invoked when file is first executed

export const siteName = globalSeo[locale as keyof typeof globalSeo].siteName;
```

becomes
```ts
export const siteName = () => {
  const locale = getLocale(); // invoked whenever this function is being called
  const localeSeo = globalSeo[locale as keyof typeof globalSeo];

  return localeSeo.siteName;
}
```

jbmoelker: The middleware ensures the locale is always configured before anything on the page happens. But the more dynamic (app-y) this setup becomes, that could become an issue. So I think this is a good change! 👏

jbmoelker (approved):

PR #205, opened 2024-11-06
feature: link to files (for download)

### LINKED ISSUE #148: Downloadable files
# User story

As a website visitor,
I want to be able to download files from the website,
so I can (store and) view them (later).

# Requirements

- Downloads have clear filenames
- Download links show meta info (language, format, size) [for accessibility and energy consciousness](https://a11y-guidelines.orange.com/en/articles/download-links/)

jbmoelker:
# Changes

- Adds a new `file` model that provides a generic pattern for content editors to add and reference files.
  - Editors can upload an asset and link to the file records in text and other fields.
  - Editors can select the locale of a file for improved accessibility.
- Adds a `LinkToFile` component. See its README for details.
- Adds a `/files/[...filename].ts` route to proxy files, so they can be used as downloads.

In CMS:
image

In Page (EN and NL):
image
image

# Associated issue

Resolves #148
Closes #147 (outdated earlier attempt based on older routing)

# How to test

1. Navigate to the [Files demo page (en)](https://feat-file-downloads.head-start.pages.dev/en/demos/file-demo/)
2. Review the links and verify that they work.
3. Switch to nl locale.
4. Review the links (see they now also display the language of the file) and verify that they work.
5. Navigate to the page in the CMS
6. Test and verify how adding a file and a link to it works

# Checklist

- ~~I have added a decision log entry if the change affects the architecture or changes a significant technology~~

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -54,7 +54,6 @@ export default defineConfig({
   output: isPreview ? 'server' : 'static',
   server: { port: localhostPort },
   site: siteUrl,
-  trailingSlash: 'always',
   vite: {
     plugins: [
       graphql() as PluginOption,
--- a/src/components/LinkToFile/LinkToFile.astro
+++ b/src/components/LinkToFile/LinkToFile.astro
@@ -4,7 +4,8 @@ import prettyBytes from 'pretty-bytes';
 import { getLocale } from '@lib/i18n';
 import { getFileHref } from '@lib/routing/';
 import Icon from '@components/Icon';
-type Props = {
+
+export type Props = {
   record: FileRouteFragment;
 };
 const { record, ...props } = Astro.props;
--- a/src/components/LinkToFile/LinkToFile.astro
+++ b/src/components/LinkToFile/LinkToFile.astro
@@ -1,0 +1,43 @@
+---
+import type { FileRouteFragment } from '@lib/types/datocms';
+import prettyBytes from 'pretty-bytes';
+import { getLocale } from '@lib/i18n';
+import { getFileHref } from '@lib/routing/';
+import Icon from '@components/Icon';
+type Props = {
+  record: FileRouteFragment;
+};
+const { record, ...props } = Astro.props;
+const { file, locale: fileLocale, title } = record;
+const pageLocale = getLocale();
+const localeText =
+  fileLocale && fileLocale !== pageLocale
+    ? new Intl.DisplayNames([pageLocale], { type: 'language' }).of(fileLocale)
+    : '';
+const format = file.format.toUpperCase();
+const size = prettyBytes(file.size, { locale: pageLocale });
+const metaText = [localeText, format, size].filter(Boolean).join(', ');
+---
+
+<a
+  href={getFileHref(record)}
+  hreflang={fileLocale}
+  download={title}
+  type={file.mimeType}
+  {...props}
+>
+  <span class="title"><slot>{title}</slot></span>
+  ({metaText})
+  <Icon class="icon" name="download" /></a
+>
+
+<style>
+  /* Underline only the title, so the meta data doesn't obscure the link visually, 
+     but is still part of the link content for assistive technology */
+  a {
+    display: contents;
+  }
+  .title {
+    text-decoration: inherit;
+  }
+</style>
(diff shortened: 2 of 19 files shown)


astro.config.ts
```diff
@@ -54,7 +54,6 @@ export default defineConfig({
   output: isPreview ? 'server' : 'static',
   server: { port: localhostPort },
   site: siteUrl,
-  trailingSlash: 'always',
```
jbmoelker, line 57: This is required for the file proxy to work (or otherwise the file href need a `/` suffix like `dummy.pdf/`) which is odd. I'm okay with having the required trailing slash removed as all our routing and canonicals make sure we have consistent page routes ending with a `/`.

jurgenbelien: Do we have tests in place to verify that trailing slashes for routes to records other than files do in fact work?

src/components/LinkToFile/LinkToFile.astro
```diff
@@ -0,0 +1,43 @@
...
+type Props = {
+  record: FileRouteFragment;
+};
+const { record, ...props } = Astro.props;
+const { file, locale: fileLocale, title } = record;
+const pageLocale = getLocale();
+const localeText =
+  fileLocale && fileLocale !== pageLocale
+    ? new Intl.DisplayNames([pageLocale], { type: 'language' }).of(fileLocale)
+    : '';
```
jurgenbelien, line 16: Awesome

src/components/LinkToRecord/LinkToRecord.astro
```diff
@@ -12,6 +13,11 @@ export type Props = {
 const { record, openInNewTab, ...props } = Astro.props;
 const locale = getLocale() as SiteLocale;
```
jurgenbelien, line 14: Hmm I should've moved the `as SiteLocale` to the `getLocale()` function itself when I had the chance

src/components/LinkToRecord/LinkToRecord.astro
```diff
@@ -12,6 +13,11 @@ export type Props = {
...
 const href = getHref({ locale, record });
+const isType = (type: string) => record.__typename === type;
 ---
 
-<Link {href} {openInNewTab} {...props}><slot>{record.title}</slot></Link>
+{
+  (isType('FileRecord'))
+    ? <LinkToFile { record } {...props}><slot>{record.title}</slot></LinkToFile>
```
jurgenbelien, line 21: This throws an error due to a number of things:
* spacing inside the braces results in typescript throwing an unhelpful type error
* `record` is of type `RecordRoute`, but `LinkToRecord` expects `FileRouteFragment`
* `FileRouteFragment` is a superset of `RecordRoute` adding `file` and `id`

Remove the function and Typescript will narrow it correctly

```suggestion
{
  (record.__typename === 'FileRecord')
    ? <LinkToFile {record} {...props}><slot>{title}</slot></LinkToFile>
```

jurgenbelien (changes requested):
Record should be the right type. I think that's of bigger benefit than having the abstraction of the function

src/components/LinkToRecord/LinkToRecord.astro
jbmoelker: @Jamerrone is about to file an issue in setting locale based on server/client context. I think that could be a good issue to address this in as well

astro.config.ts
jbmoelker: Not yet. I see the routing tests only cover redirects: https://github.com/voorhoede/head-start/blob/main/src/lib/routing/index.test.ts#L31
Will add tests

src/lib/routing/index.test.ts
```diff
@@ -1,65 +1,107 @@
 import { describe, expect, test, vi } from 'vitest';
-import { defaultRedirectStatus, getRedirectTarget, redirectStatusCode } from '@lib/routing/redirects';
```
jbmoelker, line 2: I moved these to `redirects.test.ts` as these cover what's in `redirects.ts` and not what's in `index.ts`. Did that in a separate commit hoping git would understand this 🤷 .

astro.config.ts
jbmoelker: Added tests. See commit https://github.com/voorhoede/head-start/pull/205/commits/587caf9b56a9035154297db944c7e65d2f73c4c4

jbmoelker: Figured I should actually also include tests for the LinkToFile component itself. WIP: https://github.com/voorhoede/head-start/pull/205/commits/970401f4a1d099dcb4d377f287e48c00a75619af (didn't have enough time to finish it)

PR #206, opened 2024-11-08
fix: conditionally set default locale on server & client

Jamerrone:
# Changes

- When `src/lib/i18n/index.ts` is run in the client, extract the current locale from the `lang` attribute. Otherwise, it just resets to the default locale. This fixes `t()` and `getLocale()` in combination with `client:load` components.

# How to test

- Use either functions inside a component with `client:load`. Translations should now work as expected.

# Checklist

- [X] I have performed a self-review of my own code
- [X] I have made sure that my PR is easy to review (not too big, includes comments)
- [X] I have made updated relevant documentation files (in project README, docs/, etc)
- [X] I have added a decision log entry if the change affects the architecture or changes a significant technology
- [X] I have notified a reviewer

### DIFF
--- a/src/lib/i18n/index.ts
+++ b/src/lib/i18n/index.ts
@@ -10,7 +10,7 @@ export const cookieName = 'HEAD_START_LOCALE';
 const i18n = rosetta(messages);
 
 if (typeof document !== 'undefined') {
-  i18n.locale(document.documentElement.lang as SiteLocale);
+  setLocale(document.documentElement.lang as SiteLocale);
 } else {
   i18n.locale(defaultLocale);
 }
--- a/src/lib/i18n/index.ts
+++ b/src/lib/i18n/index.ts
@@ -8,7 +8,12 @@ export const defaultLocale = locales[0];
 export const cookieName = 'HEAD_START_LOCALE';
 
 const i18n = rosetta(messages);
-i18n.locale(defaultLocale);
+
+if (typeof document !== 'undefined') {
+  i18n.locale(document.documentElement.lang as SiteLocale);
+} else {
+  i18n.locale(defaultLocale);
+}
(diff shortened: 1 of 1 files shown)


src/lib/i18n/index.ts
```diff
@@ -8,7 +8,12 @@ export const defaultLocale = locales[0];
 export const cookieName = 'HEAD_START_LOCALE';
 
 const i18n = rosetta(messages);
-i18n.locale(defaultLocale);
+
+if (typeof document !== 'undefined') {
+  i18n.locale(document.documentElement.lang as SiteLocale);
```
jbmoelker, line 13: Nice! I would use `setLocale` to protect against unsupported locales.

I'm also curious if we could add a test for this. Our `renderFragment` doesn't include the `document`, so we'll need a different method. Or maybe only mock just the `document` for this file?

jbmoelker (changes requested):

jbmoelker (approved):
Tested this by following [Astro islands tutorial](https://docs.astro.build/en/tutorial/6-islands/1/).

Replaced `Greeting.jsx` with:
```jsx
import { useState } from 'preact/hooks';
import { locales, getLocale, setLocale, t } from '@lib/i18n';

export default function Greeting({ messages }) {
  const [locale, setStateLocale] = useState(getLocale());

  const nextLocale = () => {
    const newLocale = locales[(locales.indexOf(locale) + 1) % locales.length];
    setStateLocale(newLocale);
    setLocale(newLocale);
  };

  return (
    <div>
      <h3>
        {t('select_language')} ({locale})
      </h3>
      <button onClick={() => nextLocale()}>Next locale</button>
    </div>
  );
}
```

Works like a charm!

Issue #196, opened 2024-11-01
Editable site menu


jbmoelker:
# User stories

As a website visitor,
I want to have access to an app/site menu,
so that I get an overview of the site content/structure and can directly navigate to pages.

As a content editor,
I want to be able to edit/compose the app/site menu,
so that I can guide users to specific pages.

# Requirements

- The menu items should be nestable (see example below) to support more complex menus.
- The menu items should support links to internal pages, external links, and non-link items that only serve to group sub items.
- The menu should use clear landmark (`<nav>` with `aria-label` or `aria-labelledby`).
- The menu should not require any library yet, but should be suitable to be used in a [Radix Navigation Menu[(https://www.radix-ui.com/primitives/docs/components/navigation-menu#navigation-menu) or similar in a project using Head Start.

Example menu (taking Head Start itself as an example):

- demos
  - Text block
  - Image block
  - Video block
- docs
  - Getting Started
  - Routing
    - Editable redirects 
  - Accessibility

jbmoelker:
I made a quick PoC for this feature in a clone of the No Dodos website, since it relied on the new linked record routing already in place there. My findings:

Result in the browser:

![image](https://github.com/user-attachments/assets/4090cd81-2dcf-4233-bdba-0cc977714339)

Works well to have a modular content field for `menuItems` in a singleton `app` model (model was already in No Dodos website)

![image](https://github.com/user-attachments/assets/0a5cfb04-50f1-4fec-9cab-b6880191b82a)

I used a menu item that supported nested `items` (referring to its own content block):

![image](https://github.com/user-attachments/assets/54c405f5-b074-47be-add1-31f26a472d92)

I used a single menu item for all 3 types of items, with a required `title` and optional (for grouping only item) `link` field. Where the link could be a record or an external link (newly introduced model): 

![image](https://github.com/user-attachments/assets/d8b7c20d-af54-405b-9e55-9b689dfc8ca3)

I introduced the external link model so it would be just as easy to link to an external as an internal page. But that gives a bit of an odd behaviour the next time you want to select a record to link to as suddenly that external url is an option you can select:

![image](https://github.com/user-attachments/assets/94bc16c0-9533-48f8-b43c-8893baccae2e)

I discussed this with @jurgenbelien and would now instead opt for 3 separate menu item blocks:
- menu item - internal link
- menu item - external link
- menu item - group

[...]

Issue #27, opened 2023-10-24
Seed (script) for new project


jbmoelker:
# User story

As a website developer,
I want to quickly seed content in a new project,
so that I can use that as a starting point.

# Checklist

- [ ] seeds SEO preferences (global settings)
- [ ] seeds home page
- [ ] seeds a few generic pages
- [ ] seeds the not found page
- [ ] seeds the translations
- [ ] seeds the (... extend as more functionality is added)
- [ ] add instructions to `docs/getting-started.md#create-a-datocms-project`

jbmoelker:
For the No Dodos website I wrote a quick script to seed translations, which could be used as a reference implementation:
[`nododos/nododos-website/scripts/seed-translations.ts`](https://github.com/No-Dodos/nododos-website/blob/910f3aaa4f80dd3764325be59475d5da870a9084/scripts/seed-translations.ts).

jbmoelker:
[DatoCMS suggests scripts to backup records and assets](https://www.datocms.com/docs/import-and-export/export-data). DatoCMS doesn't offer a way to import those same backups. We could use the format the records are downloaded in as a source for the seed files. This would also provide us a way to restore backups in the future (not that that's something I'm too worried about, as DatoCMS makes its own backups).

PR #230, opened 2024-12-20
feature: basic editable responsive menu

jbmoelker:
# Changes

- Adds an `AppMenu` to the default layout header. The menu items are editable via the CMS (new models included in migration). The menu has basic responsive behaviour with a full and compact mode. See `AppMenu/README.md` for details.
- Adds icons for the open and close menu buttons.
- Refactors inline logic to small components so they could be reused in multiple places (`AppLogo`, `IconButton`, `SearchLink` and `UnstyledList`).
- Moves components that should only be used in layouts to `src/layouts/`, just like the new `AppMenu`.
- Adds / updates 7 small READMEs.

Only `AppMenu` is really new. All the other changes are assets (icons), refactors and documentation. So the number of files changed may be significant. But the actual changes are not that big.

Note: other types of menu items (like external links) and nested menu items can be added in a separate PR so this one doesn't become any bigger.

# Associated issue

Part of #196.

# How to test

1. Open preview link
2. Verify the header contains the new app menu with items from the CMS
3. Verify the menu items work
4. Verify that when the menu items no longer fit, the menu items are hidden and a menu button appears.
5. Verify the menu button opens the menu in a dialog and you can operate it (focus management, tab and esc work).
6. Change the menu in the CMS
7. Verify the new menu items are rendered (requires local dev server)

# Checklist

- [x] I have made sure that my PR is easy to review (not too big, includes comments) (only `AppMenu` is really new, other files are moved around and many short `README`s are added)


src/components/UnstyledList/README.md
```diff
@@ -0,0 +1,5 @@
+# Unstyled List
+
+**Restores semantic list role on an unstyled list.**
+
+Lists require `[role="list"]` attribute when `list-style: none;` is applied to maintain accessibility in VoiceOver in Safari. See [explanation on web.dev](https://web.dev/articles/website-navigation#:~:text=The%20WebKit%20team%20decided%20to%20remove%20list%20semantics%2C%20when%20a%20list%20doesn%27t%20look%20like%20a%20list.).
```
jbmoelker, line 5: Better reference? https://developer.mozilla.org/en-US/docs/Web/CSS/list-style#accessibility

jurgenbelien: MDN's wording is a bit vague but this is only the case for lists outside of `<nav>`. Not saying that we shouldn't add role="list"regardless, just wanted to clarify.

image

src/layouts/AppMenu/AppMenu.astro
```diff
@@ -0,0 +1,113 @@
...
+    <span id="main-menu-title" class="a11y-sr-only">{t('main_menu')}</span>
+    <UnstyledList class:list="main-menu__list" data-menu-list>
+      {items.map((item) => <AppMenuItem {item} class:list="main-menu__item" />)}
+      <li class="main-menu__item main-menu__item--last"><SearchLink /></li>
+    </UnstyledList>
+  </nav>
+  <LocaleSelector pageUrls={pageUrls} />
+  <IconButton icon="menu" label={t('open_menu')} data-menu-button hidden />
+
+  <dialog data-menu-dialog>
```
jurgenbelien, line 30: No popover API? 😅

src/layouts/AppMenu/AppMenuItem.astro
```diff
@@ -0,0 +1,21 @@
+---
+import type { AppMenuFragment } from '@lib/datocms/types';
+import LinkToRecord from '@components/LinkToRecord/LinkToRecord.astro';
+
+export interface Props {
+  item: AppMenuFragment['menuItems'][0];
```
jurgenbelien, line 6:
```suggestion
  item: AppMenuFragment['menuItems'][number];
```
This does not have an impact in this case, but there might be situations where an array is interpreted as a tuple by Typescript, and if this pattern is repeated in that instance you would have incorrect types.

src/layouts/AppMenu/AppMenu.client.ts
```diff
@@ -0,0 +1,44 @@
...
+  #onResize() {
+    this.classList.remove('is-compact');
+    const isFitting = this.#menuList.scrollWidth <= this.#menuList.clientWidth;
+    this.classList.toggle('is-compact', !isFitting);
+  }
+
+  connectedCallback() {
+    this.#menuButton.removeAttribute('hidden');
+    this.#menuButton.addEventListener('click', this.open.bind(this));
+    this.#dialog.addEventListener('click', this.close.bind(this));
```
jurgenbelien, line 32: I find this behavior a bit weird. Clicking anywhere except the links in the dialog will close the menu dialog. What if I want to copy some text from the menu dialog? That would be impossible at the moment, because the dialog is dismissed the moment I release my mouse/trackpad button. I would expect the close handler to be only on the backdrop and the close button.

jurgenbelien (changes requested):
I feel that the close behavior of the menu dialog is a bit to eager in closing itself.

config/datocms/migrations/1706210000_translations.ts
```diff
@@ -72,6 +72,9 @@ export default async function (client: Client) {
     'skip_to_content': 'Skip to content',
     'not_found': 'The page you requested could not be found.',
     'watch_video_on_provider': 'watch on {{provider}}',
+    'main_menu': 'main menu',
+    'open_menu': 'open menu',
+    'close_menu': 'close menu',
```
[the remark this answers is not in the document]
jbmoelker: I think we should use seeds for this. See #27 . But this is the best we have for now for new projects using Head Start.

src/layouts/AppMenu/AppMenu.astro
jbmoelker: No, that would open as `.show()` instead of `.showModal()`. So can't use popover API :-/

src/layouts/AppMenu/AppMenu.client.ts
jbmoelker: Will do.

src/components/UnstyledList/README.md
jbmoelker: So developers can decide when to use this UnstyledList component.

src/layouts/AppMenu/AppMenu.client.ts
jbmoelker: Resolved with separate event listener for close button and check if click is outside the dialog.

src/layouts/AppMenu/AppMenu.client.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.#menuButton.focus();
+  }
+
+  #onDialogClick(event: MouseEvent) {
+    const rect = this.#dialog.getBoundingClientRect();
+    const isClickOutside = 
+      event.clientY < rect.top ||
+      event.clientY > rect.bottom ||
+      event.clientX < rect.left ||
+      event.clientX > rect.right;
```
jurgenbelien, line 31: nice

jurgenbelien (approved):

Issue #10, opened 2023-08-11
CMS editor experience


jbmoelker:
# User story

As a content editor,
I want a CMS that supports me in managing content autonomously,
so that I can do my job confidently and effectively.

# Checklist

- [ ] Fields have help texts, so I understand what they are for
- [ ] Modular content has preview images, so I know what I’m selecting -> #313 
- [ ] Sidebar and views are well organised (incl emoji), so I can quickly find content
- [x] Content models have preview links, so I can review before publishing
- [ ] CMS has customer’s theme (logo, colours), so it feels familiar to me
- [ ] Custom help page ([plugin](https://www.datocms.com/marketplace/plugins/i/datocms-plugin-custom-page)), to help new editors get started (incl recording for example)

Issue #170, opened 2024-09-09
Translation plugin on text fields


jbmoelker:
As a content editor,
I want to automatically translate text content in the CMS,
so that I can do my job more efficiently.

# Checklist

- [ ] Add [Voorhoede translations plugin](https://www.datocms.com/marketplace/plugins/i/datocms-plugin-translate-fields) to all relevant text fields
- [ ] Document how to configure API key (either manually, or using an env variable?)

PR #233, opened 2024-12-21
feature: preview links in CMS

jbmoelker:
# To do

- Get preview secret to be used automatically. If click on the preview links in the sidebar you still need to enter the secret manually (I don't get why). While if you copy the link and paste it in a new browser tab it works as expected.

# Changes

This change contains 2 solutions:

- an `/api/reroute/page` endpoint to find and redirect to the correct page url
- a check in `pages/[locale]/[...path]/index.astro` that redirects to the canonical url if it doesn't match the current url

To do: agree on one, remove the other.

# Associated issue

Part of #10 

# How to test

1. Open the CMS and go to the `preview-links` environment (note: it's moved from personal to De Voorhoede organisation account, so we could add more sandbox environments)
2. Open the sidebar in the home / page / not found page model
3. Try the preview links
4. Verify that they all work as expected

# Checklist

### DIFF
--- a/config/datocms/migrations/1734806654_previewLinks.ts
+++ b/config/datocms/migrations/1734806654_previewLinks.ts
@@ -1,0 +1,64 @@
+import { Client } from '@datocms/cli/lib/cma-client-node';
+
+export default async function (client: Client) {
+  console.log('Manage upload filters');
+
+  console.log('Install plugin "Model Deployment Links"');
+  await client.plugins.create({
+    id: 'MKba9NT5QBKZaeI4HcERwA',
+    package_name: 'datocms-plugin-model-deployment-links',
+  });
+  await client.plugins.update('MKba9NT5QBKZaeI4HcERwA', {
+    parameters: { datoApiToken: process.env.DATOCMS_PREVIEW_API_TOKEN },
+  });
+
+  console.log('Creating new fields/fieldsets');
+
+  console.log(
+    'Create JSON field "Preview" (`preview`) in model "\uD83D\uDCD1 Page" (`page`)'
+  );
+  await client.fields.create('LjXdkuCdQxCFT4hv8_ayew', {
+    id: 'XF2LuFVWSrmu7Lle8xlhTg',
+    label: 'Preview',
+    field_type: 'json',
+    api_key: 'preview',
+    localized: true,
+    appearance: {
+      addons: [],
+      editor: 'MKba9NT5QBKZaeI4HcERwA',
+      parameters: { urlPattern: '/{ locale }/{ slug }/' },
+    },
+  });
+
+  console.log(
+    'Create JSON field "Preview" (`preview`) in model "\uD83C\uDFE0 Home" (`home_page`)'
+  );
+  await client.fields.create('X_tZn3TxQY28ltSyjZUGHQ', {
+    id: 'YPXZOMoWRdKTHLUkN9ytfw',
+    label: 'Preview',
+    field_type: 'json',
+    api_key: 'preview',
+    localized: true,
+    appearance: {
+      addons: [],
+      editor: 'MKba9NT5QBKZaeI4HcERwA',
+      parameters: { urlPattern: '/{ locale }/' },
+    },
+  });
+
+  console.log(
+    'Create JSON field "Preview" (`preview`) in model "\uD83E\uDD37 Not found" (`not_found_page`)'
+  );
+  await client.fields.create('d_AvMVoMSqmNbMqx-NdqIw', {
+    id: 'O_DCfpaDSzq1MX4DaRlMpQ',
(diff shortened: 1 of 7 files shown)


src/pages/api/reroute/page.ts
```diff
@@ -0,0 +1,38 @@
...
+const jsonResponse = (data: object, status: number = 200) => {
+  return new Response(JSON.stringify(data), {
+    status,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  });
+};
+
+export const GET: APIRoute = async ({ request }) => {
```
jbmoelker, line 18: This is a generic alternate approach that I used in the nododos-website. But I think the canonical url check in `[locale]/[...path]/index.astro` is easier and probably the way to go.

config/datocms/migrations/1734806654_previewLinks.ts
```diff
@@ -0,0 +1,64 @@
...
+export default async function (client: Client) {
+  console.log('Manage upload filters');
+
+  console.log('Install plugin "Model Deployment Links"');
+  await client.plugins.create({
+    id: 'MKba9NT5QBKZaeI4HcERwA',
+    package_name: 'datocms-plugin-model-deployment-links',
+  });
+  await client.plugins.update('MKba9NT5QBKZaeI4HcERwA', {
+    parameters: { datoApiToken: process.env.DATOCMS_PREVIEW_API_TOKEN },
```
jbmoelker, line 12: I'm struggling where to define this token. Similar issue as with the #170. Maybe rather add it to the getting started documentation that these tokens need to be created and configured in the CMS?
(this is why currently all the CI checks are failing).

src/pages/404.astro
```diff
@@ -15,14 +12,16 @@ export const prerender = false;
 
 Astro.response.status = 404;
 
-const { locale } = Astro.params as { locale: SiteLocale };
+const localeFromPath = Astro.params.locale as SiteLocale;
+const localeFromQuery = Astro.url.searchParams.get('locale') as SiteLocale;
+const locale = localeFromQuery || localeFromPath;
```
jbmoelker, line 17: This is now needed to have the `Astro.rewrite('/404/')` work in `[locale]/[...path]/index.astro`. It's the tradeoff for no longer needing to cast the response from the page query. Worth it? Or rather have that one fail as it did before?

src/pages/[locale]/[...path]/index.astro
```diff
@@ -53,19 +53,30 @@ type Params = {
 
 const { locale, path } = Astro.params as Params;
 const variables = { locale, slug: getPageSlugFromPath(path) };
-const { page } = (await datocmsRequest<PageQuery>({ query, variables })) as {
-  page: NonNullable<PageQuery['page']>; // Only NonNullable when statically generated. Handle as a 404 when this is a server route!
-};
-const breadcrumbs = [...getParentPages(page), page].map((page) =>
+const { page } = await datocmsRequest<PageQuery>({ query, variables });
+
+if (!page) {
```
Comment (jbmoelker) [inline src/pages/[locale]/[...path]/index.astro:58]: This check ensures this type casting is no longer needed:
```ts
as {
  page: NonNullable<PageQuery['page']>; // Only NonNullable when statically generated. Handle as a 404 when this is a server route!
};
```

But it makes the `404` rewrite and locale setting in that file more complex. Worth it?

config/datocms/migrations/1734806654_previewLinks.ts
jbmoelker: Decided to handle this in the getting started documentation instead.

docs/getting-started.md
```diff
@@ -66,6 +66,18 @@ You can now run your project locally:
 npm run dev
 ```
 
+### Configure DatoCMS plugins
+
+Head Start comes with a few DatoCMS plugins pre-installed. Some plugins require additional project-specific configuration:
+
+#### Model Deployments Links plugin
+
+The [Model Deployment Links plugin](https://www.datocms.com/marketplace/plugins/i/datocms-plugin-model-deployment-links) requires an API key:
```
jbmoelker, line 75: Note: this could also be achieved within a migration script using https://www.datocms.com/docs/content-management-api/resources/access-token/create . However it's difficult to reliably get the Editor role ID. Since this is a one time config step, I'm happy to have it as a manual step, rather than scripting the generation of access token.

src/pages/api/reroute/page.ts
jbmoelker: Hmmm. I guess the canonical check works on localhost and preview as the generic page route is then always handled dynamically on the server. However the production environment has prerendered pages and will throw a 404 I suppose. So I guess we do need this reroute API handler.

In that case the config of the Preview field of the generic page will need to use that one instead.

PR #235, opened 2024-12-28
feature: HTML to Markdown (component & routes)

jbmoelker:
# Changes

Making our HTML pages available as Markdown is of interest to some people and especially suitable for software such as LLM's. These changes make HTML content available as Markdown.

- Adds a `ToMarkdown` component to render HTML as Markdown (works both run- and build-time).
- Adds a `[locale]/[...path]/index.md.astro` route to render matching route as Markdown (works both run- and build-time).
- Adds post build script to rename Markdown routes to `.md` as Astro always generates `.md/index.html` files.
- Adds support for Github Flavoured Markdown, including tables.

See decision log entry for background.

# Associated issue

N/A

# How to test

1. Open preview link
2. Navigate to a page, like `/en/documentation/getting-started/`
3. Replace the trailing slash with `.md`, like `/en/documentation/getting-started.md`
4. Verify the page is now rendered as Markdown.
5. Navigate to a page with tables, like `/en/demos/tables-demo.md`
6. Verify that tables are rendered to Markdown.

# Checklist


src/pages/[locale]/[...path]/index.md.astro
```diff
@@ -0,0 +1,16 @@
+---
+import ToMarkdown from '@components/ToMarkdown/ToMarkdown.astro';
+import Page from './index.astro';
+export { getStaticPaths } from './_index';
+
+export const partial = true;
+
+Astro.response.headers.set('Content-Type', 'text/markdown; charset=utf-8');
```
Comment (jbmoelker) [inline src/pages/[locale]/[...path]/index.md.astro:8]: HTML entities in the content are still encoded (`<` becomes `&lt;` etc). This is not the cause of the `ToMarkdown` component as it also happens with the HTML comment below. So Astro forces this higher up I believe. Haven't been able to stop encoding or force decoding 🤷 .

src/components/ToMarkdown/README.md
```diff
@@ -0,0 +1,19 @@
+# To Markdown
+
+**Renders nested HTML content to Markdown.**
+
+## Examples
```
jbmoelker, line 5: I wanted to write these and others as test. But while this component works in a route, I'm not able to isolate the output properly. 🤷 .

src/pages/[locale]/[...path]/_index.ts
```diff
@@ -0,0 +1,32 @@
+import { datocmsCollection } from '@lib/datocms';
+import { type PageRouteForPath, getPagePath } from '@lib/routing/page';
+
+export async function getStaticPaths() {
```
Comment (jbmoelker) [inline src/pages/[locale]/[...path]/_index.ts:4]: Could keep this in `index.astro` and try to import it in `index.md.astro`. I like this separate file as it makes it more clear it can be used across routes. And it also obfuscated the content of `index.astro` a bit. So I'm happy with this.

src/pages/[locale]/[...path]/index.md.astro
```diff
@@ -0,0 +1,16 @@
+---
+import ToMarkdown from '@components/ToMarkdown/ToMarkdown.astro';
+import Page from './index.astro';
```
Comment (jbmoelker) [inline src/pages/[locale]/[...path]/index.md.astro:3]: I like this pattern where the `Page` from `index.astro` can remain unmodified and this route does all the `ToMarkdown` magic. If a developer using Head Start doesn't need this behaviour, this `index.md.astro` route can simply be removed.

docs/decision-log/2024-12-27-render-to-markdown.md
```diff
@@ -0,0 +1,31 @@
...
+
+**Use `Astro.slots.render` to transform HTML content to Markdown both during build- and run-time.**
+
+- Date: 2024-12-27
+- Alternatives Considered: `AstroContainer.renderToString`, to Markdown per block, post build script to transform HTML to Markdown.
+- Decision Made By: [Jasper](https://github.com/jbmoelker)
+
+## Decision
+
+Making our HTML pages available as Markdown is of interest to some people and especially suitable for software such as LLM's. Using `Astro.slots.render` seems to be the only solution that works both during build- and run-time.
```
jurgenbelien, line 11:
```suggestion
Making our HTML pages available as Markdown is of interest to some people and especially suitable for software such as LLMs. Using `Astro.slots.render` seems to be the only solution that works both during build- and run-time.
```

docs/decision-log/2024-12-27-render-to-markdown.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Solution
+
+A componanion `.md.astro` route with a `ToMarkdown` component that leverages `Astro.slots.render` to get a usable HTML string, which can be used to transform to Markdown (and other formats):
```
jurgenbelien, line 15:
```suggestion
A companion `.md.astro` route with a `ToMarkdown` component that leverages `Astro.slots.render` to get a usable HTML string, which can be used to transform to Markdown (and other formats):
```

jurgenbelien (commented):
It works (apart from the html entities) and there are some changes that are IMHO beneficial to the project regardless (moving staticPaths to separate file). 

I do wonder if this functionality makes sense out of the box, instead of adding it later on for a project that needs it. For starters this increases build time for every project, even if they don't use it. A developer might remove the markdown route but forget the link tag, serving erroneous alternates that might negatively impact SEO.

jbmoelker:
Will be replaced by #345

PR #242, opened 2025-01-06
feature: Service worker (custom Astro integration)

### LINKED ISSUE #199: Add Service Worker
# User story

As a website visitor,
I want web pages to be cached,
so I get a fast and reliable experience.

# Strategy

Add a Service Worker with a strategy to use recently cached pages (x minutes ago?), then try the network (if older or not in cache), then cache (if network fails, and an older cached page is at least still available).

We could use [Workbox](https://developer.chrome.com/docs/workbox/modules/workbox-strategies) for this.

decrek:
# Changes

- Add service worker setup with basic network first strategy on HTML pages
- Made the decision to go for my own custom integration. It can serve as an example integration and I like the fact that we build our own service worker from a service-worker.ts file instead of generating one. I feel this is the most understandable for others and gives you maximum freedom over its implementation.

# Associated issue

Resolves #199 

# How to test

1. Open preview link
2. Open devtools --> Application
3. Verify a service worker is installed
4. Go offline, refresh the page and verify that the HTML page is still served

# Checklist

### DIFF
--- a/config/astro/service-worker-integration.ts
+++ b/config/astro/service-worker-integration.ts
@@ -7,6 +7,16 @@ const filenamePath = (filename: string) => fileURLToPath(new URL(join('../../', 
 const srcFilename = filenamePath('src/assets/service-worker.ts');
 const outFilename = filenamePath('dist/service-worker.js');
 
+const buildConfig = {
+  entryPoints: [srcFilename],
+  outfile: outFilename,
+  target: ['es2020'],
+  bundle: true,
+  minify: true,
+  allowOverwrite: true,
+  sourcemap: true,
+};
+
 export default function serviceWorkerIntegration(): AstroIntegration {
   return {
     name: 'service-worker',
@@ -28,13 +38,8 @@ export default function serviceWorkerIntegration(): AstroIntegration {
       'astro:build:done': async () => {
         try {
           await esbuild.build({
-            entryPoints: [srcFilename],
-            outfile: outFilename,
-            target: ['es2020'],
-            bundle: true,
-            minify: true,
-            allowOverwrite: true,
-            sourcemap: true,
+            ...buildConfig,
+            write: true,
           });
         } catch (e) {
           console.error('Failed to build service worker');
@@ -48,13 +53,9 @@ export default function serviceWorkerIntegration(): AstroIntegration {
 
 export const GET: APIRoute = async () => {
   const output =  await esbuild.build({
-    entryPoints: [srcFilename],
-    outdir: outFilename,
-    target: ['es2020'],
-    bundle: true,
-    minify: false,
+    ...buildConfig,
     write: false,
-    allowOverwrite: true,
+    minify: false,
     sourcemap: false,
   });
 
--- a/config/astro/service-worker-integration.ts
+++ b/config/astro/service-worker-integration.ts
@@ -2,8 +2,8 @@ import type { APIRoute, AstroIntegration } from 'astro';
 import { fileURLToPath } from 'node:url';
 import * as esbuild from 'esbuild';
 
-const serviceWorkerSrc = fileURLToPath(new URL('../src/assets/service-worker.ts', import.meta.url));
-const serviceWorkerDist = fileURLToPath(new URL('../dist/service-worker.js', import.meta.url));
+const serviceWorkerSrc = fileURLToPath(new URL('../../src/assets/service-worker.ts', import.meta.url));
+const serviceWorkerDist = fileURLToPath(new URL('../../dist/service-worker.js', import.meta.url));
 
 export default function serviceWorkerIntegration(): AstroIntegration {
   return {
@@ -37,6 +37,7 @@ export default function serviceWorkerIntegration(): AstroIntegration {
         } catch (e) {
           console.error('Failed to build service worker');
           console.error(e);
+          process.exit(1);
         }
       },
     },
--- a/config/astro/service-worker-integration.ts
+++ b/config/astro/service-worker-integration.ts
@@ -1,9 +1,11 @@
 import type { APIRoute, AstroIntegration } from 'astro';
+import { join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import * as esbuild from 'esbuild';
 
-const serviceWorkerSrc = fileURLToPath(new URL('../../src/assets/service-worker.ts', import.meta.url));
-const serviceWorkerDist = fileURLToPath(new URL('../../dist/service-worker.js', import.meta.url));
+const filenamePath = (filename: string) => fileURLToPath(new URL(join('../../', filename), import.meta.url));
+const srcFilename = filenamePath('src/assets/service-worker.ts');
+const outFilename = filenamePath('dist/service-worker.js');
 
 export default function serviceWorkerIntegration(): AstroIntegration {
   return {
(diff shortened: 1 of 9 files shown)


scripts/service-worker-integration.ts
```diff
@@ -0,0 +1,64 @@
+import type { APIRoute, AstroIntegration } from 'astro';
```
jbmoelker, line 1: I think I would move this file to `config/astro/service-worker-integration.ts` or `config/astro/integrations/service-worker.ts`, since it's basically a configuration snippet that's loaded into the main `astro.config.ts`. Whereas `/scripts/` contains all scripts that are executed from a cli context.

scripts/service-worker-integration.ts
```diff
@@ -0,0 +1,64 @@
...
+import { fileURLToPath } from 'node:url';
+import * as esbuild from 'esbuild';
+
+const serviceWorkerSrc = fileURLToPath(new URL('../src/assets/service-worker.ts', import.meta.url));
```
jbmoelker, line 5: I think within an Astro integration you have access to the configured src and dist directory (`srcDir` and `outDir` on the `config` param) ([reference](https://github.com/withastro/astro/blob/main/packages/integrations/sitemap/src/index.ts#L70-L78)).

So I would change the API of the `serviceWorkerIntegration` so it can be called as something like:

```ts
serviceWorkerIntegration({
  srcFilename: 'assets/service-worker.ts', // relative to `srcDir`,
  outFilename: 'service-worker.ts',        // relative to `outDir`,
})
```

src/assets/service-worker.ts
```diff
@@ -0,0 +1,31 @@
+/// <reference lib="webworker" />
+import { registerRoute } from 'workbox-routing';
+import { NetworkFirst } from 'workbox-strategies';
+import { CacheableResponsePlugin } from 'workbox-cacheable-response';
+import { ExpirationPlugin } from 'workbox-expiration';
+
+((self: ServiceWorkerGlobalScope) => {
```
jbmoelker, line 7: Maybe this can be simplified by [declaring `self`](https://www.devextent.com/create-service-worker-typescript/)?:

```ts
/// <reference lib="webworker" />
declare const self: ServiceWorkerGlobalScope;
```

src/components/PerfHead/PerfHead.astro
```diff
@@ -20,3 +20,17 @@ const fontFaceDeclaration = fonts.map((font) => getFontFaceDeclaration(font)).jo
   ))
 }
 <style set:html={fontFaceDeclaration}></style>
+
+<script is:inline>
```
jbmoelker, line 24: PerfHead sounds like the perfect location for this :)

src/assets/service-worker.ts
jurgenbelien: I looked at this with @decrek and there are two ways to do it:

```
declare const self: ServiceWorkerGlobalScope;
export default null // or export default {}
```
or do the self-invoking function to assign it to a new similarly named variable.

Without the export statement, it'll throw errors.

scripts/service-worker-integration.ts
```diff
@@ -0,0 +1,64 @@
...
+        } catch (e) {
+          console.error('Failed to build service worker');
+          console.error(e);
+        }
+      },
+    },
+  };
+}
+
+export const GET: APIRoute = async () => {
```
decrek, line 46: Lets include the asset manifest, see https://docs.astro.build/en/reference/integrations-reference/#astrobuilddone

decrek (commented):
Shall we

src/assets/service-worker.ts
decrek: 🆗

scripts/service-worker-integration.ts
jbmoelker: We tried. Had issues hoisting to root scope. Slightly refactored instead.

config/astro/service-worker-integration.ts
```diff
@@ -0,0 +1,67 @@
...
+          console.error(e);
+          process.exit(1);
+        }
+      },
+    },
+  };
+}
+
+export const GET: APIRoute = async () => {
+  const output =  await esbuild.build({
```
jbmoelker, line 50: The `esbuild.build` config between dev and prod are (and should be) mostly the same (only `minify` and `sourcemap` are different). Should we move the shared config up to the root?

```ts
const buildConfig = {
    entryPoints: [srcFilename],
    outdir: outFilename,
    target: ['es2020'],
    bundle: true,
    write: false,
    allowOverwrite: true,
}
```

and use that in both places to keep everything in sync?, like:

```ts
const output = await esbuild.build({
   ...buildConfig,
   minify: false,
   sourcemaps: false,
});
```

src/assets/service-worker.ts
```diff
@@ -0,0 +1,32 @@
...
+});
+
+registerRoute(
+  ({ request }) => request.mode === 'navigate',
+  new NetworkFirst({
+    networkTimeoutSeconds: 3,
+    cacheName: 'pages',
+    plugins: [
+      new CacheableResponsePlugin({
+        statuses: [0, 200],
```
jbmoelker, line 25: Should we consider a package like [`http-status-codes`](https://www.npmjs.com/package/http-status-codes) to write `StatusCodes.OK` instead of `200`? Or simply use some `const`s at the top, like `statusOpaque = 0;` and `statusOk = 200`? Or does this only complicate things?

(ironically that package doesn't have the opaque response status; I also had to look that up. So maybe a comment or link on that would be helpful instead)

src/assets/service-worker.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ExpirationPlugin } from 'workbox-expiration';
+
+self.addEventListener('install', () => {
+  self.skipWaiting();
+});
+  
+self.addEventListener('activate', (event) => {
+  event.waitUntil(self.clients.claim());
+});
+
```
jbmoelker, line 17: We discussed caching assets as well. Do we want to add that to this PR? Something like:

```ts
registerRoute(
  ({ request }) => request.url.pathname.startsWith('/_astro/'),
  new CacheFirst({
    cacheName: 'assets',
    plugins: [
      new CacheableResponsePlugin({
        statuses: [0, 200],
      }),
      new ExpirationPlugin({
        maxEntries: 50,
        maxAgeSeconds: 30 * 24 * 60 * 60, // 30 days
      }),
    ],
  })
);
```
What would be good expiration settings?

Or should we match `immutable` responses instead? Like:

```ts
registerRoute(
  async ({ request, response }) => {
    if (!response) return false;
    const cacheControl = response.headers.get('cache-control');
    return cacheControl?.split(' ').includes('immutable') ?? false;
  },
  /* ... same */
);
```

jbmoelker (approved):
Looks good and works well! I pushed some basic documentation and added comments with a few more suggestions.

src/assets/service-worker.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference lib="webworker" />
+declare const self: ServiceWorkerGlobalScope;
+export default null;
+
+import { registerRoute } from 'workbox-routing';
+import { NetworkFirst } from 'workbox-strategies';
+import { CacheableResponsePlugin } from 'workbox-cacheable-response';
+import { ExpirationPlugin } from 'workbox-expiration';
+
```
jbmoelker, line 9: Should we add config that makes cache cleanup easier in the future? Like:

```ts
import { setCacheNameDetails } from 'workbox-core';
import pkg from '@root/package.json';

setCacheNameDetails({
  prefix: pkg.name,
  suffix: 'v1',
});
```

PR #246, opened 2025-01-11
feature: migration helper scripts + docs

### LINKED ISSUE #211: Document migrations setup
Should include:

- using environments
- creating / autogenerating migrations (could add a `migrations:create` script to help out)
- use maintenance mode (could use script `npx datocms maintenance:on|off`, see [datocms docs](https://www.datocms.com/docs/scripting-migrations/apply-migrations-to-primary-environment))
- apply migrations (could use a `migrations:apply` script to help out)
- promoting environments (could use a `migrations:promote` script to help out, see [promote](https://www.datocms.com/docs/content-management-api/resources/environment/promote))
- refer to this in `docs/upgrading.md` as performing a migration is required in case of new migration files.

jbmoelker:
Resolves #211

# Changes

# Associated issue

# How to test

# Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -1,0 +1,10 @@
+# Migrations
+
+**Head Start uses DatoCMS migrations to develop CMS changes in isolation and ensure they can be reproduced.**
+
+
+Run
+
+```shell
+npm run cms:environment:create
+```
--- a/scripts/cms-apply-migration.ts
+++ b/scripts/cms-apply-migration.ts
@@ -1,0 +1,4 @@
+import { execCommand } from './lib/exec-command';
+import { datocmsEnvironment } from '../datocms-environment';
+
+execCommand(`npx datocms migrations:run --destination=${datocmsEnvironment} --fast-fork`);
--- a/scripts/cms-create-environment.ts
+++ b/scripts/cms-create-environment.ts
@@ -1,0 +1,21 @@
+import { buildClient } from '@datocms/cma-client-node';
+import dotenv from 'dotenv-safe';
+import { execCommand } from './lib/exec-command';
+import { datocmsEnvironment } from '../datocms-environment';
+
+dotenv.config({
+  allowEmptyValues: Boolean(process.env.CI),
+});
+
+const getPrimaryEnvironment = async () => {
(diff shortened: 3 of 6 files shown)


scripts/cms-run-migrations.ts
```diff
@@ -0,0 +1,8 @@
+import { execCommand } from './lib/exec-command';
+import { datocmsEnvironment } from '../datocms-environment';
+
+const confirmationMessage = `Apply all migrations in chronological order to the environment '${datocmsEnvironment}'`;
+execCommand(
```
velomovies, line 5: What about an option to run a migration `--in-place`?
Let a developer select if it needs to be in-place or with creating a new env.
```js
if (migrationType === 'in-place') {
    return await execCommandSafe(
      `npx datocms migrations:run --source=${targetEnvironment} --in-place`,
    );
  } else {
    return await execCommandSafe(
      ` npx datocms migrations:run --destination=${targetEnvironment}`,
    );
  }

**Review by velomovies — COMMENTED** [review]
For 'create', 'destroy' and 'run' a good option would be to add the ability to add a custom environment name in the CI
```js
export const getTargetEnvironment = async () => {
  const choice = await select({
    message: 'What sandbox environment do you want to target?',
    choices: [
      {
        name: `Use default: '${datocmsEnvironment}'`,
        value: 'default',
        description: 'Use the environment specified in datocms-environment.ts',
      },
      {
        name: 'Enter custom name',
        value: 'custom',
        description: 'Specify a custom environment name',
      },
    ],
  });

  if (choice === 'custom') {
    const customEnv = await input({
      message: '📝 Enter the environment name:',
      validate: (value) => {
        if (!value.trim()) {
          return 'Environment name cannot be empty';
        }
        return true;
      },
    });
    return customEnv;
  }

  return datocmsEnvironment;
};
```
Implementation of @MarleenEliza

scripts/cms-run-migrations.ts
jurgenbelien: I definitely see the need to add an option for `--in-place` but I'd rather put it in a separate script and try to fit it in the defined workflow.

jurgenbelien (approved):
Tested flow incl promoting and destroying environments. I think the manage scripts are very helpful

PR #250, opened 2025-01-16
Fix Video Blocks in non Chromium browsers

### LINKED ISSUE #249: VideoBlock uses navigator.connection which might not be available in every browser
The VideoBlock component uses `#useDataSaveMode` in [src/blocks/VideoBlock/Video.client.ts](https://github.com/voorhoede/head-start/blob/main/src/blocks/VideoBlock/VideoBlock.client.ts) to find out if data saving is needed. `navigation.connection` however is not available in Edge, Firefox, and Safari.

```ts
#useDataSaveMode = (() => {
  const navigator = (window.navigator as Navigator & { connection?: { saveData: boolean } })
  return navigator.connection?.saveData || false; 
})()
```

decrek:
# Changes

add feature check for navigator.connection since it is chromium only making video blocks fail outside of chromium browsers

# Associated issue

Resolves #249 

# How to test

Test Video Block in non Chromium browser

# Checklist

### DIFF
--- a/src/blocks/VideoBlock/VideoBlock.client.ts
+++ b/src/blocks/VideoBlock/VideoBlock.client.ts
@@ -26,6 +26,10 @@ class VideoBlock extends HTMLElement {
    * @see https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation
    */
   #useDataSaveMode = (() => {
+    if (!('connection' in navigator)) {
+      return false;
+    }
+    
     type NetworkInformation = { saveData: boolean; }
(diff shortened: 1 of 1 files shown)


jurgenbelien (approved):
I would personally write the whole function as:

```ts
#useDataSaveMode = (() => {
  const navigator = window.navigator as Navigator & { connection?: { saveData: boolean } }
  return navigator.connection?.saveData || false; 
})()
```
But the outcome is the same.

jbmoelker (approved):
❤️

PR #254, opened 2025-03-10
refactor: Type datocmsRequest based on output mode or prerender.

jurgenbelien:
In order to cut down on type assertion via `as`, `datocmsRequest` should return a more workable type definition. Based on a passed type variable – or when omitted, the value of `output` in the astro config – `datocmsRequest` either has a return type that assumes the resource is available, or passes the expected return value based on the query.

# Changes

- Export output mode from astro config.
- Allow asserted type override of datocmsRequest via `AssertReturnType`. 

# How to test

1. Check out code locally
2. Open `@pages/[locale]/[...path]/index.astro`
3. Verify that the return value of `datocmsRequest` is typed by `PageQuery`, without any type assertion.
4. Replace comment `/* typeof prerender */` with a `true` or `false` value in the `datocmsRequest` type variables. 
5. Verify that any mention of `page` gets TypeScript errors.
6. Remove the manipulated type varable.
7. See that the TypeScript errors disappear.
8. Change the output mode in `config/output.ts`;
5. Verify that the TypeScript errors return.

# Checklist

### DIFF
--- a/config/output.ts
+++ b/config/output.ts
@@ -1,0 +1,13 @@
+import { type AstroUserConfig } from 'astro';
+
+/**
+ * Output mode for the Astro build process.
+ * - 'static': Pre-renders (SSG) all pages at build time into HTML files (default).
+ * - 'server': Use server-side rendering (SSR) for all pages by default, 
+ *    always outputting a server-rendered site.
+ * - The 'server' mode is used when preview mode is enabled, allowing dynamic SSR.
+ * Overriding individual pages to use SSR or SSG is possible by adding 
+ * `export const prerender = true // or false;` to a route.
+ * @see https://docs.astro.build/en/reference/configuration-reference/#output
+ */
+export const output = 'static' as const satisfies AstroUserConfig['output'];
--- a/src/lib/datocms/index.ts
+++ b/src/lib/datocms/index.ts
@@ -2,7 +2,8 @@ import { Kind, parse, type DocumentNode, type FragmentDefinitionNode } from 'gra
 import { print } from 'graphql/language/printer';
 import type { SiteLocale } from '@lib/i18n/types';
 import { titleSuffix } from '@lib/seo';
-import { datocmsBuildTriggerId, datocmsEnvironment } from '../../../datocms-environment';
+import { datocmsBuildTriggerId, datocmsEnvironment } from '@root/datocms-environment';
+import { output } from '@root/config/output';
 import { DATOCMS_READONLY_API_TOKEN, HEAD_START_PREVIEW } from 'astro:env/server';
 
 const wait = (milliSeconds: number) => new Promise((resolve) => setTimeout(resolve, milliSeconds));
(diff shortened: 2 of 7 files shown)


src/lib/datocms/index.ts
```diff
@@ -16,12 +17,51 @@ type DatocmsRequest = {
   variables?: { [key: string]: string };
   retryCount?: number;
 };
+
+
+/**
+ * Expect return value of specified Query to be available, i.e. non-nullable.
+ */
+type Assumed<Query> = Required<{ [key in keyof Query]: NonNullable<Query[key]> }>;
```
jurgenbelien, line 25: Consider for instance `PageQuery`:
```
type PageQuery = { 
  __typename?: 'Query', 
  page?: { 
    __typename: 'PageRecord', 
   id: string,
  // etc...
  } | null 
}
```

* `Required<...>` removes the  `?:` on all top level properties. 
*  `[key in keyof Query]` loops over those properties.
* `NonNullable<...>` removes the null or undefined from each of the properties.

src/lib/datocms/index.ts
```diff
@@ -16,12 +17,51 @@ type DatocmsRequest = {
...
+): Promise<Assumed<Query>>;
+export async function datocmsRequest<Query>(
+  options: DatocmsRequest & { prerender: false }
+): Promise<Query>;
+export async function datocmsRequest<
+  Query,
+  Prerender extends boolean = typeof output extends 'static' ? true : false
+>(
+  options: DatocmsRequest & { prerender?: Prerender }
+): Promise<Prerender extends true ? Assumed<Query> : Query>;
```
jurgenbelien, line 53: These are all the overloads for the `datocmsRequest` function, explicitly stating the return value based on the `prerender` value. The last one makes sure that the default value of the generic `Prerender` is based on the value of `output`.

src/lib/datocms/index.ts
```diff
@@ -73,13 +113,13 @@ type CollectionMeta = {
  * This function uses pagination to get all records.
  * @see https://www.datocms.com/docs/content-delivery-api/pagination
  */
-export const datocmsCollection = async <CollectionType>({
+export async function datocmsCollection<CollectionType>({
```
jurgenbelien, line 116: For consistency's sake I made this a regular function as well.

config/output.ts
jurgenbelien, line 1: This needs to be a separate file in order to import it into `@lib/datocms/index.ts`

config/output.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Output mode for the Astro build process.
+ * - 'static': Pre-renders (SSG) all pages at build time into HTML files (default).
+ * - 'server': Use server-side rendering (SSR) for all pages by default, 
+ *    always outputting a server-rendered site.
+ * - The 'server' mode is used when preview mode is enabled, allowing dynamic SSR.
+ * Overriding individual pages to use SSR or SSG is possible by adding 
+ * `export const prerender = true // or false;` to a route.
+ * @see https://docs.astro.build/en/reference/configuration-reference/#output
+ */
+export const output = 'static' as const satisfies AstroUserConfig['output'];
```
jurgenbelien, line 13: `as const` so that `datocmsRequest` has an inferred default.
`satisfies` so that the options remain clear.

See that the Typescript hint is the actual string 'static'

sjoerdbeentjes: When setting this to 'server', [this line ](https://github.com/voorhoede/head-start/pull/254/files#diff-536daf7ee02131ca906a17a9304fe8e0a7b53d135d47733667624d3c0ebeae57R61) breaks. Not exactly sure why and how to fix it though.

sjoerdbeentjes (changes requested):
Looks good! Will definitely save some frustration

config/output.ts
jurgenbelien: I'll have a look!

src/lib/datocms/index.ts
```diff
@@ -16,12 +17,50 @@ type DatocmsRequest = {
...
+): Promise<Prerender extends true ? Assumed<Query> : Query>;
+export async function datocmsRequest<
+  Query,
+  Prerender extends boolean = typeof output extends 'static' ? true : false
+>(
+  {
+    query,
+    variables = {},
+    retryCount = 1,
+    prerender = (output as OutputModes === 'static') as Prerender,
```
jurgenbelien, line 61: By widening the type of `output` with `as OutputModes`, the typing of this line is still correct, even when `output` would be set to `server`. All other definitions are still inferred from the value of `output` itself.

src/lib/datocms/index.ts
```diff
@@ -16,12 +17,37 @@ type DatocmsRequest = {
   variables?: { [key: string]: string };
   retryCount?: number;
 };
+
+/**
+ * Expect return value of specified Query to be available, i.e. non-nullable.
+ */
+type Assert<Query> = Required<{ [key in keyof Query]: NonNullable<Query[key]> }>;
```
jurgenbelien, line 24: Consider for instance `PageQuery`:
```
type PageQuery = { 
  __typename?: 'Query', 
  page?: { 
    __typename: 'PageRecord', 
   id: string,
  // etc...
  } | null 
}
```

* `Required<...>` removes the  `?:` on all top level properties. 
*  `[key in keyof Query]` loops over those properties.
* `NonNullable<...>` removes the null or undefined from each of the properties.

(This is a new comment after I changed the name of this type)

src/lib/datocms/index.ts
```diff
@@ -16,12 +17,37 @@ type DatocmsRequest = {
...
+ * @param options.query - The GraphQL query document
+ * @param options.variables - Variables to pass to the GraphQL query
+ * @param options.retryCount - Number of retry attempts in case of rate limiting
+ * 
+ * @returns A promise resolving to the query result, with optional type assertion
  */
-export const datocmsRequest = async <T>({ query, variables = {}, retryCount = 1 }: DatocmsRequest): Promise<T> => {
+export async function datocmsRequest<
+  Query,
+  AssertReturnValue extends boolean = typeof output extends 'static' ? true : false
```
jurgenbelien, line 43: Infer the default behavior of route rendering from the value of output (`as const` helps by getting this as a type).

jurgenbelien:
Had a discussion with @decrek and this could very well be an issue we should address in the types generated from GraphQL, for instance by a generic type to assert properties of the return values.

sjoerdbeentjes (approved):
Looking & working great 🚀

Issue #256, opened 2025-03-12
Split page partial into component to group content and component to reuse content


jurgenbelien:

jbmoelker:
Suggestion: Create a separate TabsBlock and AccordionBlock as that's what editors best understand. The PagePartialBlock can keep doing what it does, a part of a page that can be reused on multiple pages.

jurgenbelien:
While I agree that having the layout in the name is more accessible to content editors, I think we should still stick to the `GroupingItem` to collect all allowed blocks.

PR #245, opened 2025-01-08
fix: use dynamic IDs in migrations

bazottie:
Fixes several previous migrations relying on non-existent hardcoded IDs when creating a new project.
Updating previous migrations is not supposed to happen, however, no migration logic is changed.
It now relies on the API to find the right IDs using the corresponding label.

## How to reproduce
1. Follow "Getting started guide" from `main` on a personal DatoCMS account
1. Upon running `npx datocms migrations:run --destination=start --fast-fork` at the corresponding step, expect an error

Example of a related error 
(saved during debugging, this is not the first error to show up, but they are all similar)
```bash
Running migration "1730405111_linkRecords.ts"... failed!

----
Error: PUT https://site-api.datocms.com/item-types/EkWwcYfPQSGXskAZKO_tTg: 404 Not Found

[
  {
    "id": "302048",
    "type": "api_error",
    "attributes": {
      "code": "NOT_FOUND",
      "details": {}
    }
  }
]
```

I inferred the itemTypes for `page` and `home_page` from inspecting an existing DatoCMS instance that uses Head Start, please check if this is correct.

# Changes
- Updates references to non existent IDs with API calls to find the correct ID in a new instance
- Adds error handling for finding `menuItems` since it now relies on finding IDs in a list

# How to test

1. Remove the start environment if it was created when attempting to reproduce the error
    1. __If testing without reproducing, simply follow getting started guide__
1. Run `npx datocms migrations:run --destination=start --fast-fork`
1. Migrations should succeed

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have made sure that my PR is easy to review (not too big, includes comments)
- [x] I have made updated relevant documentation files (in project README, docs/, etc)
- [x] I have added a decision log entry if the change affects the architecture or changes a significant technology
- [x] I have notified a reviewer

PR #257, opened 2025-03-13
feature: add grouping block

MarleenEliza:
# Changes

- based on #256 , adds a `GroupingBlock` with `GroupingItemBlock`. 

This is meant to replace a part of the `Page Partial Block` functions, but since the editing of the `Page Partial Block` is a breaking change that will impact already existing `Page Partial Block`s, this change will be dealt with in a different PR.
This PR only **adds** a `GroupingBlock` which is a non-breaking change.

# Associated issue

#256 (non breaking half)

# How to test

# Checklist


src/blocks/GroupingBlock/GroupingBlock.test.ts
```diff
@@ -0,0 +1,242 @@
+import { renderToFragment } from '@lib/renderer';
+import { describe, expect, test } from 'vitest';
+import GroupingBlock, { type Props } from './GroupingBlock.astro';
+
+describe('GroupingBlock', () => {
+  test('renders a stack layout without a title when layout is "stack-untitled"', async () => {
```
MarleenEliza, line 6: Same layout tests as done previously on Page Partial

src/blocks/GroupingBlock/GroupingBlock.astro
```diff
@@ -0,0 +1,61 @@
...
+import type { AnyBlock } from '@blocks/Blocks';
+
+export interface Props {
+    block: GroupingBlockFragment
+}
+
+type GroupingItem = {
+  title: string;
+  blocks: AnyBlock[];
+};
```
MarleenEliza, line 15: `GroupingItem` is only used inside GroupingBlock and does not exsist as a separate Block the content editor can add.

MarleenEliza (commented):
@jurgenbelien Could you have a look at this PR? It is non breaking and add the `GroupingBlock` only

src/blocks/GroupingBlock/GroupingItemBlock/GroupingItemBlock.fragment.graphql
jurgenbelien, line 1: I think we can move this out of it's containing folder, and put the whole fragment definition inside of the other block's fragment file

src/blocks/GroupingBlock/GroupingBlock.test.ts
jurgenbelien: Let's not duplicate it. Remove the PagePartial tests that are about rendering it with a certain layout, and let the PagePartialBlock use the GroupingBlock to render.

src/blocks/GroupingBlock/GroupingBlock.astro
jurgenbelien: Note that you do have a Fragment for this that you can import...

config/datocms/migrations/1741855116_addGroupingBlock.ts
```diff
@@ -0,0 +1,224 @@
...
+          'BRbU6VwTRgmG5SbwUs0rBg',
+          'F60ZY1wFSW2qbvh99poj3g',
+          'PAk40zGjQJCcDXXPgygUrA',
+          'QYfZyBzIRWKxA1MinIR0aQ',
+          'TBuD6qQOSDy6i9dM3T_XEA',
+          'VZvVfu52RZK81WG0Dxp-FQ',
+          'V80liDVtRC-UYgd3Sm-dXg',
+          'ZdBokLsWRgKKjHrKeJzdpw',
+          'gezG9nO7SfaiWcWnp-HNqw',
+          '0SxYNS2CR1it_5LHYWuEQg',
```
jurgenbelien, line 198: Consider writing this in a nondestructive way. See https://github.com/voorhoede/head-start/pull/245

For this migration I'd say do the following for pages and homepage:

```
const homePageBodyBlocks = (await client.fields.find("home_page::body_blocks"))
  .validators?.rich_text_blocks?.item_types || [];

await client.fields.update('pUj2PObgTyC-8X4lvZLMBA', {
  validators: {
    rich_text_blocks: {
      item_types: [
        ...homePageBodyBlocks,
        //  grouping block.
```

jurgenbelien (commented):
Let's make sure the grouping logic is only in grouping block, and PagePartial uses the logic in GroupingBlock to render it (for the time being)

config/datocms/migrations/1741855116_addGroupingBlock.ts
```diff
@@ -0,0 +1,224 @@
...
+      inverse_relationships_enabled: false,
+    },
+    {
+      skip_menu_item_creation: true,
+      schema_menu_item_id: 'XZDLSGZcSvGjkpzdfymBrg',
+    },
+  );
+
+  console.log(
+    'Create block model "\uD83D\uDDC2\uFE0F Grouping Item Block" (`grouping_item_block`)',
```
jurgenbelien, line 25: Should we maybe name it `GroupingItem` instead of `GroupingItemBlock`? It should only be used inside the `GroupingBlock`, and should not be in `AnyBlock`.

src/blocks/PagePartialBlock/PagePartialBlock.astro
```diff
@@ -1,60 +1,21 @@
...
+---
+<!-- Temporary solution to prevent breaking changes for ecsisting Page Partial Blocks-->
+<GroupingBlock block={{
+  ...block,
+  __typename: 'GroupingBlockRecord',
+  items: block.items.map(item => ({
+    title: item.title,
+    blocks: item.blocks as GroupingBlockFragment['items'][number]['blocks'],
+    __typename: 'GroupingItemBlockRecord'
+  }))
```
MarleenEliza, line 20: Not a very elegant way of typing, but mind that this is temporary

src/blocks/GroupingBlock/GroupingBlock.astro
```diff
@@ -0,0 +1,59 @@
...
+import Blocks from '@blocks/Blocks.astro';
+import { Accordion, AccordionItem } from '@components/Accordion';
+import { Tabs, TabsTab, TabsPanel } from '@components/Tabs';
+import type { AnyBlock } from '@blocks/Blocks';
+
+export interface Props {
+    block: GroupingBlockFragment
+}
+export type GroupingItem = GroupingItemBlockFragment & {
+    blocks: AnyBlock[];
```
jurgenbelien, line 12: I would not expect `AnyBlock` here, the definition of blocks is in `GroupItem(Block)Fragment`

src/blocks/GroupingBlock/GroupingBlock.fragment.graphql
```diff
@@ -0,0 +1,12 @@
+#import './GroupingItemBlock.fragment.graphql'
```
jurgenbelien, line 1: Paste the contents of this fragment underneath the `GroupingBlock` fragment

src/blocks/GroupingBlock/GroupingBlock.astro
jurgenbelien: Just wanted to follow up on this, because turns out this is a bit more complicated than I anticipated and I want to make sure that we know why:

Consider the type of AnyBlock: it is a union of all defined fragments. Such a fragment might look like this:
```ts
export type ActionBlockFragment = {
  __typename: 'ActionBlockRecord',
  id: string,
  items: Array<...>
}
export type ImageBlockFragment = {
  __typename?: 'ImageBlockRecord', // not sure why this one is undefined, I copy/pasted it from `datocms/types.ts`
  id: string
  image: { ... }
};
```

Now one might think passing a subset of `AnyBlock` to `Blocks.astro` as props would be fine, because that's what the union is for, right? 
Unfortunately, we're doing something else every time we're listing blocks, which is actually ruining this for us, and that is including `__typename` in the list of blocks. This makes it possible to emit to the console that we've not implemented a block renderer correctly, but it also means that these are the types generated for it:

```ts
blocks: Array<
    {
      __typename: 'ActionBlockRecord'
      id: string
      items: Array< ... >
    } | { 
      __typename: 'ImageBlockRecord' 
    } 
```

This means that any block that is not allowed wil still be present as a `__typename: 'SomethingRecord'` type. And, of course,  `{ __typename: 'ImageBlockRecord' }`, is not assignable to type ` { __typename: 'ImageBlockRecord', id: string, image { ... } }`. 

## Now what?
If we want to maintain the `__typename` runtime hint behavior, I think we need to loosen `AnyBlock`:

```ts
export type AnyBlock = Omit<
  (
    | ActionBlockFragment
    | EmbedBlockFragment
    | ImageBlockFragment
    | PagePartialBlockFragment
    | TableBlockFragment
    | TextBlockFragment
    | TextImageBlockFragment
    | VideoBlockFragment
    | VideoEmbedBlockFragment
    | GroupingBlockFragment 
  ), 
  '__typename'
> & { __typename: string };
```

@jbmoelker @decrek Are there any reasons not to do this?

jurgenbelien: https://github.com/voorhoede/head-start/pull/259

config/datocms/migrations/1741855116_addGroupingBlock.ts
```diff
@@ -1,12 +1,20 @@
 import { Client } from '@datocms/cli/lib/cma-client-node';
 
 export default async function (client: Client) {
+  
+  //@ts-expect-error rich_text_blocks is only available on Modular Content fields
+  const homeBodyBlocks = (await client.fields.find('home_page::body_blocks')).validators.rich_text_blocks?.item_types;
+  //@ts-expect-error rich_text_blocks is only available on Modular Content fields
+  const pageBodyBlocks = (await client.fields.find('page::body_blocks')).validators.rich_text_blocks?.item_types;
+  //@ts-expect-error rich_text_blocks is only available on Modular Content fields
+  const pagePartialBlocks = (await client.fields.find('page_partial::blocks')).validators.rich_text_blocks?.item_types;
```
MarleenEliza, line 10: Nice to see that you can get everything easily from the client!

_I don't fully understand why it's inside `rich_text_blocks` though; is this how DatoCMS structures fields inside blocks?_
The above was what I was originally planning on asking, but I see that DatoCMS indeed uses this structure when creating blocks too!

```
await client.fields.create(groupingItem.id, {
   ...
    validators: {
      rich_text_blocks: {
        item_types: [
```

MarleenEliza (commented):
Looking good! Nice to see it was easy to get the `pagePartialBlocks` from Dato. 

I don't know if it is relevant for now, but I just remembered we implemented a `startingTab` of sorts for DGTL. Might be a nice feature to have in the Grouping Block in the future.
As well as a 'grid' of some sorts.

PR #262, opened 2025-04-01
fix(seo): also return seo settings when project has one locale

petergoes:
Add helper function to get proper locale data.

When there is only one locale configured in Dato, the `seoGlobal` is populated
with all data directly instead of with a locale key in between

### DIFF
--- a/src/lib/seo/index.ts
+++ b/src/lib/seo/index.ts
@@ -9,19 +9,19 @@ export type PageUrl = {
   pathname: string,
 };
 
-export const siteName = () => {
+/** 
+  * `globalSeo` _should_ have a key per available locale. When there is only one
+  * locale configured in Dato, that key is missing. Therefore we fallback to the
+  * `globalSeo` object 
+  */
+const localeSeo = () => {
   const locale = getLocale();
-  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
-
-  return localeSeo.siteName;
+  const localeSeoData = globalSeo[locale as keyof typeof globalSeo];
+  return localeSeoData || globalSeo;
 };
 
-export const titleSuffix = () => {
-  const locale = getLocale();
-  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
-
-  return localeSeo.titleSuffix;
-};
+export const siteName: () => string = () => localeSeo().siteName;
+export const titleSuffix: () => string = () => localeSeo().titleSuffix;
 
 export const noIndexTag: Tag = {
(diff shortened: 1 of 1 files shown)


src/lib/seo/index.ts
```diff
@@ -9,19 +9,19 @@ export type PageUrl = {
...
 };
 
-export const titleSuffix = () => {
-  const locale = getLocale();
-  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
-
-  return localeSeo.titleSuffix;
-};
+export const siteName: () => string = () => localeSeo().siteName;
+export const titleSuffix: () => string = () => localeSeo().titleSuffix;
```
jurgenbelien, line 24:
```suggestion
export const siteName = () => localeSeo().siteName;
export const titleSuffix = () => localeSeo().titleSuffix;
```
I'd rather lean on type inference here than have the explicit type for these arrow functions.

src/lib/seo/index.ts
```diff
@@ -9,19 +9,19 @@ export type PageUrl = {
...
+  * locale configured in Dato, that key is missing. Therefore we fallback to the
+  * `globalSeo` object 
+  */
+const localeSeo = () => {
   const locale = getLocale();
-  const localeSeo = globalSeo[locale as keyof typeof globalSeo];
-
-  return localeSeo.siteName;
+  const localeSeoData = globalSeo[locale as keyof typeof globalSeo];
+  return localeSeoData || globalSeo;
```
jurgenbelien, line 20: Yes this is nitpicky 😅, but I find variables that kind of hold the same value, and then differ only slightly in name quite confusing (ie. `localeSeo` and `localeSeoData`). 

```suggestion
  return globalSeo[locale as keyof typeof globalSeo] || globalSeo;
```

jurgenbelien (commented):
Good find! And I think this indeed is the most elegant solution at the moment. 
Forgive me for my nitpicky points regarding code style

jurgenbelien (approved):

src/lib/seo/index.ts
petergoes: I understand the confusion between variable names that differ slightly, however, my original line 19 was exactly that, the _locale_ version of the seo data. and it differs from the globalSeo conceptually.

the line `return localeSeoData || globalSeo` tells me the _intent_ of the code: use the locale version first, fallback to the global when not available. Which is what this function is all about. Besides, my localeSeoData name explains the (quite confusing) type magic.

Your suggestion does not convey the intent or the typescript magic.

What do you think?

PR #276, opened 2025-06-25
F

luukbrauckmann:
# Changes

- Add migration script for Plugin Phone Number

# Associated issue

# How to test

# Checklist

### DIFF
--- a/config/datocms/migrations/1750855236_phoneNumberPlugin.ts
+++ b/config/datocms/migrations/1750855236_phoneNumberPlugin.ts
@@ -1,0 +1,25 @@
+import { Client } from '@datocms/cli/lib/cma-client-node';
+
+export default async function (client: Client) {
+  console.log('Manage upload filters');
+
+  console.log('Install plugin "Phone Number"');
+  await client.plugins.create({
+    id: 'CAQilGMyRlGtygeH_obytw',
+    package_name: 'datocms-plugin-phone-number',
+  });
+
+  console.log('Update existing fields/fieldsets');
+
+  console.log(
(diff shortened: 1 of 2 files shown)


config/datocms/migrations/1750855236_phoneNumberPlugin.ts
```diff
@@ -0,0 +1,25 @@
...
+    id: 'CAQilGMyRlGtygeH_obytw',
+    package_name: 'datocms-plugin-phone-number',
+  });
+
+  console.log('Update existing fields/fieldsets');
+
+  console.log(
+    'Update Single-line string field "Phone number" (`phone_number`) in block model "\uD83D\uDCDE Phone Link" (`phone_link`)',
+  );
+  await client.fields.update('bfnOOp5cSM-x8S5RKEpKsw', {
```
jurgenbelien, line 17: Could you set up this update so that it does not depend on the id of the field. Instead, look up the field like so:

  `const PhoneNumberField = (await client.fields.find('phone_link::phone_number')`
  
  Make sure that you only update the settings that are set by the plugin (ie, editor/parameters and I'm guessing field_extension?

jurgenbelien (commented):
Please make sure the migration will never clash when cherry-picking this commit in other projects

config/datocms/migrations/1750855236_phoneNumberPlugin.ts
luukbrauckmann: I implemented the `find()` like you suggested. I've also removed the id setting for the plugin and instead let the migration handle it self. I've also re-applied the field's appearance settings, so no settings will be accidentally overwritten.

PR #279, opened 2025-06-30
Add prompt flow  

MarleenEliza:
# Changes

- adds Docs for migration workflow
- adds `cms:manage` option to help with choosing the right prompt
- Changes workflow slightly to use names that align more with Headstart use cases instead of DatoCMS terminology.

# Associated issue

# How to test

# Checklist


scripts/lib/environments.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const getPrimaryEnvironment = async () => {
+  const client = buildClient({ apiToken: process.env.DATOCMS_API_TOKEN });
+  const environments = await client.environments.list();
+  const primaryEnvironment = environments.find(env => env.meta.primary)?.id;
+  if (!primaryEnvironment) {
+    throw new Error('No primary environment found');
+  }
+  return primaryEnvironment;
+};
```
jurgenbelien, line 72:
```suggestion
export const getPrimaryEnvironment = async () => {
  const primary = await execAsync('npx datocms environments:primary')
    .then(({ stdout }) => stdout.trim());
    
    if (!primary) {
      throw new Error('No primary environment found');
    }
    
    return primary
};
```

PR #282, opened 2025-07-04
feature: page partials content collection

### LINKED ISSUE #166: Content collections
As a website visitor,
I want content pages organised in collections,
so I can easily find similar / related content

* A content editor must be able to organise pages in collections (or tags?)
* A Collection Route must use dynamic (server-side) rendering, so that it supports pagination, filtering (incl searching) and sorting
* A Collection Block can show items from a collection (or multiple) with filters and sorting applied and a limit on the number of items. A layout property can define its appearance

jurgenbelien:
# Changes

- Adds initial implementation of content collections
- Uses content collections for page partials to reduce query complexity
- Handles localization within content collections
- Bypasses astro collections in development and preview
- Cleans up SiteLocale by using as const instead of enum for generated DatoCMS types

# Associated issue

Partially resolves #166

# How to test

1. Open preview link
2. Navigate to Page Partial Demo
4. Verify that Page Partials show up

# Checklist

### DIFF
--- a/src/blocks/PagePartialBlock/PagePartialBlock.astro
+++ b/src/blocks/PagePartialBlock/PagePartialBlock.astro
@@ -2,8 +2,7 @@
 import type { PagePartialBlockFragment } from '@lib/datocms/types';
 import GroupingBlock from '@blocks/GroupingBlock/GroupingBlock.astro';
 import type { GroupingBlockFragment } from '@lib/datocms/types';
-import { getEntry } from 'astro:content';
-import { getLocale } from '@lib/i18n';
+import { getEntry } from '@lib/content';
 
 export interface Props {
   block: PagePartialBlockFragment
@@ -13,7 +12,7 @@ const { block } = Astro.props;
 // TODO: only allow one page partial per block
 const pagePartials = await Promise.all(
   block.items.map(async (item) => {
-    const pagePartial = await getEntry('PagePartials', `${getLocale()}/${item.id}`);
+    const pagePartial = await getEntry('PagePartials', item.id);
     return pagePartial?.data;
   })
 ).then(items => items.filter(pagePartial => pagePartial !== undefined && pagePartial !== null));
--- a/src/blocks/PagePartialBlock/PagePartialBlock.astro
+++ b/src/blocks/PagePartialBlock/PagePartialBlock.astro
@@ -2,18 +2,27 @@
 import type { PagePartialBlockFragment } from '@lib/datocms/types';
 import GroupingBlock from '@blocks/GroupingBlock/GroupingBlock.astro';
 import type { GroupingBlockFragment } from '@lib/datocms/types';
+import { getEntry } from 'astro:content';
+import { getLocale } from '@lib/i18n';
 
 export interface Props {
   block: PagePartialBlockFragment
 }
 const { block } = Astro.props;
 
+// TODO: only allow one page partial per block
+const pagePartials = await Promise.all(
+  block.items.map(async (item) => {
+    const pagePartial = await getEntry('PagePartials', `${getLocale()}/${item.id}`);
+    return pagePartial?.data;
+  })
+).then(items => items.filter(pagePartial => pagePartial !== undefined && pagePartial !== null));
 ---
 {/* Temporary solution to prevent breaking changes for existing Page Partial Blocks */}
 <GroupingBlock block={{
   ...block,
   __typename: 'GroupingBlockRecord',
-  items: block.items.map(item => ({
+  items: pagePartials.map(item => ({
     title: item.title,
     blocks: item.blocks as GroupingBlockFragment['items'][number]['blocks'],
     __typename: 'GroupingItemRecord'
--- a/src/content/PagePartials/index.ts
+++ b/src/content/PagePartials/index.ts
@@ -5,6 +5,7 @@ import {
   type SiteLocale
 } from '@lib/datocms/types';
 import { datocmsCollection, datocmsRequest } from '@lib/datocms';
+import { combine } from '@lib/content';
 
 export type PagePartialCollectionEntry = PagePartialCollectionEntryQuery['entry'] & {
   recordId: string,
@@ -22,7 +23,7 @@ const loadEntry = async (id: string, locale?: SiteLocale | null) => {
   return {
     ...entry,
     recordId: entry.id,
-    id: `${locale}/${entry.id}`,
+    id: combine({ id: entry.id, locale }),
     locale,
   } satisfies PagePartialCollectionEntry;
 };
--- a/src/content/PagePartials/index.ts
+++ b/src/content/PagePartials/index.ts
@@ -1,0 +1,75 @@
+import { defineCollection, z } from 'astro:content';
+import {
+  PagePartialCollectionEntry as query,
+  type PagePartialCollectionEntryQuery,
+  type SiteLocale
+} from '@lib/datocms/types';
+import { datocmsCollection, datocmsRequest } from '@lib/datocms';
+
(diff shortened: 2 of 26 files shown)


src/content/PagePartials/index.ts
```diff
@@ -0,0 +1,57 @@
+import { defineCollection, z } from 'astro:content';
+import {
+  PagePartialCollectionEntry as query,
+  type PagePartialCollectionEntryQuery,
+  SiteLocale
```
jbmoelker, line 5: `type SiteLocale`

src/content/PagePartials/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    _allBlocksLocales: { locale: SiteLocale }[] 
+  }>({
+    collection: name,
+    fragment: /* graphql */`
+      id
+      _allBlocksLocales {
+        locale
+      }
+    `,
+  })).flatMap(({ id, _allBlocksLocales }) => _allBlocksLocales.map(({ locale }) => ({ id, locale })));
```
jbmoelker, line 39: Maybe add an explanation inline why this is needed? Or on top of the function add add some JSDoc with an example of the return value? I feel like the abstractions and mappings that we're introducing in this file might be hard to understand for the next developer (I'm already struggling a bit and we've discussed this before).

src/content/config.ts
```diff
@@ -0,0 +1,13 @@
+// Astro expects a content/config.ts file, therefore we export the collections here.
```
jbmoelker, line 1: Add a `docs/content-collections.md` (or somewhere else) to explain this whole new concept that where introducing? (or in a separate PR, also with plop files, etc)

src/content/config.ts
```diff
@@ -0,0 +1,13 @@
...
+import PagePartialCollection from './PagePartials';
+
+export const collectionMap = {
+  ...PagePartialCollection,
+} as const;
+
+export const collections = Object.entries(collectionMap).reduce<
```
jbmoelker, line 8: Why was it again that we do:
```ts
export default {
  [name]: {
    name,
    collection,
    ...
```
Instead of:
```ts
export default {
  name,
  collection,
  ...
```

It seems this file would be more readable if this could simply be:
```ts
export const allCollections = [
  PagePartialCollection,
] as const;

export const collections = allCollections.reduce((acc, { name, collection }) => {
  acc[name] = collection;
  return acc;
}, {});
```

Is it only because of `@lib/content/index.ts`? Would we consider using `.find()` there instead?

src/content/PagePartials/index.ts
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const name = 'PagePartials' as const;
+
+const loadEntry = async ({ id, locale }: { id: string, locale: SiteLocale }) => {
+  const { entry } = await datocmsRequest<PagePartialCollectionEntryQuery>({ query, variables: { id, locale } });
+  return {
+    ...entry,
+    recordId: entry.id,
+    id: `${locale}/${entry.id}`,
```
jbmoelker, line 22: Should this use the `combine()` function from `@lib/content/index.ts`?

src/lib/content/index.ts
```diff
@@ -0,0 +1,74 @@
...
+) {
+  const entry = await getAstroCollectionEntry(collection, combine(id, locale));
+  
+  // If the entry is not found, try to fetch it without locale
+  return (locale)
+    ? entry || await getAstroCollectionEntry(collection, id)
+    : entry;
+}
+
+export function combine(id: string, locale?: string | null) {
```
jbmoelker, line 59: - `split` returns an object. Maybe use the same signature here?
- Should `locale` be a `SiteLocale | null`?

src/lib/content/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export async function getEntry<Key extends CollectionKey>(
+  collection: Key,
+  id: string,
+  locale: SiteLocale | null = getLocale(),
+) {
+  const entry = await getAstroCollectionEntry(collection, combine(id, locale));
+  
+  // If the entry is not found, try to fetch it without locale
+  return (locale)
```
jbmoelker, line 54: I'm getting lost here. Maybe refactor to some return early statements?
```ts
if (entry) return entry;
// ...
```

jbmoelker: And can `locale` still be `null` at this point? The function parameter sets it to the default locale with `getLocale()`.

src/lib/content/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @param filter - Optional SiteLocale or function to filter the collection entries
+ * @returns A promise that resolves to an array of normalized collection entries
+ */
+export async function getCollection<Key extends CollectionKey>(
+  collection: Key,
+  filter?: ((entry: CollectionEntry<Key>) => boolean),
+  locale: SiteLocale | null = getLocale(),
+): Promise<CollectionEntry<Key>[]> {
+  // Filter on locale and passed filter function
+  const filterFunction = (!filter && !locale)
```
jbmoelker, line 23: Same as below, I'm getting a bit lost here. Maybe use return early statements here as well?
```ts
if (!filter && !locale) {
  return getAstroCollection(collection);
}
// ...
```

src/lib/content/index.ts
jurgenbelien: Now that the live fetching is in there it hopefully makes a bit more sense. 

`locale` can definitely be `null`, that is kind of the point of the retry for the nonlocalized lookup. You might not want to repeat yourself too much with passing the localization every time you fetch the content, but if a collection is not localized (perhaps a collection of app settings?), you might not want to prefix those localizations. A default param is only used if the param position is `undefined`, not `null`:

```ts
function defaultParams(value: string | null = 'example') {
  return value;
}

defaultParams();
// returns 'example'

defaultParams(null);
// returns null
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

jbmoelker (approved):

src/blocks/PagePartialBlock/PagePartialBlock.astro
```diff
@@ -2,18 +2,26 @@
...
 }
 const { block } = Astro.props;
 
+// TODO: only allow one page partial per block
+const pagePartials = await Promise.all(
+  block.items.map(async (item) => {
+    const pagePartial = await getEntry('PagePartials', item.id);
+    return pagePartial?.data;
+  })
+).then(items => items.filter(pagePartial => pagePartial !== undefined && pagePartial !== null));
```
jurgenbelien, line 18: Ideally a PagePartialBlock only allows one PagePartial. Since that would be a big change, we're stil working on that, trying to find a way to write a migration that will move multiple pagePartials into a grouping block of page partials

src/content/PagePartials/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    `,
+  }))
+    // Flatten the array of entries to get an array of { id, locale } pairs.
+    .flatMap(({ id, _allBlocksLocales }) => _allBlocksLocales
+      .map(({ locale }) => ({ id, locale })) 
+    );
+
+  // For each id/locale pair, load the entry and return it.
+  // Note that this might be slow if there are many entries, as it makes a
+  // separate request for each entry.
```
sjoerdbeentjes, line 58: What do you think of using grahpql query batching here? tested it and it seems faster (don't know if there are limits on how much queries you can batch, but might be worth looking into):

With:
image

Without
image


Not that much of a difference, but can imagine that this will improve things with a large number of queries

jurgenbelien: Will keep this in mind for the next interation, thanks! :)

PR #285, opened 2025-07-07
feature: Search Form Block

### LINKED ISSUE #165: Search block

jbmoelker:
# Changes

- Adds a content block that allows editors to add a search form to any page.
- The search form is connected to the search page.
- The search input is pre-filled with a query from the CMS / search param / path name. 
- Adds search form block's README.
- Order of what is filled in in the search form is -> 1. query from dato, 2. query from url, 3. path from url, 4. nothing. 

# Associated issue

Resolves #165

# How to test

1. Open preview link
2. Navigate to a page that does not exist: `/en/not-found/anywhere`.
7. Verify that the Search Form block use the path name if there is no query in the url or filled in from dato.
8. Change the URL to include a `query` search parameter, like: `/en/not-found?query=some-text`.
9. Verify that the other search form inputs use this `query` value.
10. Verify that the Search Form block can contain the query from dato and overwrites the query and path from the url.
11. Verify that you can use multiple search form blocks on one page.
12. Verify that deeplinks in the searchlinks still work. Example: search result has a link under it, which goes to a certain part of the page with the use of #

# Checklist

- ~~I have added a decision log entry if the change affects the architecture or changes a significant technology~~

### DIFF
--- a/config/datocms/migrations/1779356356_searchFormBlock.ts
+++ b/config/datocms/migrations/1779356356_searchFormBlock.ts
@@ -48,6 +48,7 @@ export default async function (client: Client) {
     label: 'Query',
     field_type: 'string',
     api_key: 'query',
+    hint: 'Pre-fill priority for the search input: 1. CMS query (this field), 2. URL query parameter (?query=…), 3. last URL path segment, 4. empty.',
     appearance: {
       addons: [],
       editor: 'single_line',
--- a/datocms-environment.ts
+++ b/datocms-environment.ts
@@ -2,5 +2,5 @@
  * @see docs/getting-started.md on how to use this file
  * @see docs/decision-log/2023-10-24-datocms-env-file.md on why file is preferred over env vars
  */
-export const datocmsEnvironment = 'deploy-table-of-contents';
+export const datocmsEnvironment = 'test-search-form-block';
 export const datocmsBuildTriggerId = '34548';
--- a/datocms-environment.ts
+++ b/datocms-environment.ts
@@ -2,5 +2,5 @@
  * @see docs/getting-started.md on how to use this file
  * @see docs/decision-log/2023-10-24-datocms-env-file.md on why file is preferred over env vars
  */
-export const datocmsEnvironment = 'test-search-form-block';
+export const datocmsEnvironment = 'deploy-search-form';
 export const datocmsBuildTriggerId = '34548';
--- a/src/blocks/SearchFormBlock/README.md
+++ b/src/blocks/SearchFormBlock/README.md
@@ -1,0 +1,11 @@
+# Search Form Block
+
+**Content block with a pre-filled search form.**
+
+## Features
+
+- Semantic search form: `form[role="search"]`.
+- Semantic search input: `input[type="search"]`.
+- Input is prefilled with query defined in CMS (`block: { query: 'text' }`) if provided.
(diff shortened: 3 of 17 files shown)


config/datocms/migrations/1779356356_searchFormBlock.ts
```diff
@@ -0,0 +1,160 @@
...
+    label: 'Query',
+    field_type: 'string',
+    api_key: 'query',
+    appearance: {
+      addons: [],
+      editor: 'single_line',
+      parameters: { heading: false, placeholder: null },
+    },
+    default_value: null,
+  });
```
MarleenEliza, line 58: Could you add a hint here as to how to fill this in? 
I know it has the following order due to Jaspers comment, but it would be nice to have this in the field description as well.

> Order of what is filled in in the search form is -> 1. query from dato, 2. query from url, 3. path from url, 4. nothing.

config/datocms/migrations/1779356356_searchFormBlock.ts
MarleenEliza, line 1: Could you rename this file to the branch name or something that indicates this is for the SearchFormBlock?

datocms-environment.ts
```diff
@@ -2,5 +2,5 @@
  * @see docs/getting-started.md on how to use this file
  * @see docs/decision-log/2023-10-24-datocms-env-file.md on why file is preferred over env vars
  */
-export const datocmsEnvironment = 'deploy-table-of-contents';
+export const datocmsEnvironment = 'test-search-form-block';
```
MarleenEliza: I will take care of the deploying / promoting to the correct env in the end , so you can ignore this change.

src/blocks/SearchFormBlock/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+**Content block with a pre-filled search form.**
+
+## Features
+
+- Semantic search form: `form[role="search"]`.
+- Semantic search input: `input[type="search"]`.
+- Input is prefilled with query defined in CMS (`block: { query: 'text' }`) if provided.
+- Input is prefilled with query from URL (`?query=text`) if otherwise provided.
+- Input is prefilled with path name from URL `/:locale/:path-name/` if no other value is available.
```
MarleenEliza: I think the explanation on the usage of the query part could be a bit clearer. 
The above is not very human-friendly imo. i know the order is
> Order of what is filled in in the search form is -> 1. query from dato, 2. query from url, 3. path from url, 4. nothing.
but that is a bit confusing when reading the README.

Also,  maybe some more concrete examples of queries would help too?

MarleenEliza (changes requested):
Looking good!
Coudl you just have a look at some of the comments I made? They are documentation orientated details, but I think they'll help with other devs understanding the component and maintenance in the future.

datocms-environment.ts
JopMolenaar: Revert this change?

src/blocks/SearchFormBlock/README.md
JopMolenaar: I updated the text. When this gets merged we can run seed docs and make a demo page for this block where users can play directly with the order of the query in this block.

MarleenEliza (approved):
Looks good! Thanks for the document revisions, I think it's easier to understand now

PR #286, opened 2025-07-09
fix: set sameSite property to 'lax' to make cookie work when navigating from external site

sjoerdbeentjes:
# Changes

Set `sameSite: 'lax'` for the preview mode cookie, otherwise it will not work when linked from a link on a page of a different domain. This is currently happening when linking to the preview mode url from Dato, where it always prompts for the secret, even when it's included in the url.

# How to test

Enter the preview mode endpoint with the configured secret from a link on a page and see if it directly goes into preview mode instead of prompting for the secret.

# Checklist

### DIFF
--- a/src/pages/api/preview/enter.ts
+++ b/src/pages/api/preview/enter.ts
@@ -7,9 +7,11 @@ export const prerender = false;
 export const cookiePath = '/';
 
 export const GET: APIRoute = async ({ cookies, locals, request }) => {
-
   if (!locals.previewSecret) {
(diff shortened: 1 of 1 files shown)


src/pages/api/preview/enter.ts
```diff
@@ -18,15 +20,15 @@ export const GET: APIRoute = async ({ cookies, locals, request }) => {
       httpOnly: true,
       secure: PUBLIC_IS_PRODUCTION,
       path: cookiePath,
-      sameSite: 'strict',
```
sjoerdbeentjes, line 21: this is what it's all about

jbmoelker (approved):

PR #293, opened 2025-07-17
Switch nvm specific config for tool agnostic version config

Siilwyn:
See https://github.com/shadowspawn/node-version-usage for tooling support.
I went with the full Node.js version, though could argue we always want the latest patch. A small downside to partial version numbers is that not all tools support this though most do...


jurgenbelien (approved):
👍

PR #294, opened 2025-07-20
feat: special purpose pages content collection

### LINKED ISSUE #291: Special purpose pages
Some pages in a website have a specific purpose while they could otherwise use the generic Page model. By introducing a mechanism for these specific purpose pages to use the generic Page model, we can prevent having to introduce and maintain specific models for each. Examples of these pages:

- **Home page**: identical to generic page, except that it is required, unique, and has no slug.
- **Search page**: identical to generic page, except that it is required, unique, and is server-rendered.
- **Not found page (404)**: identical to generic page, except that it is required, unique, has no slug, is possibly server-rendered, and should not be indexed.
- **Server error page (50x)**: identical to generic page, except that it is required, unique, has no slug, is possibly server-rendered, and should not be indexed.
- **Contact page**: currently not in scope, but typically a generic page that is specifically linked to from different places like the main menu.

I see two possible mechanisms to use generic Page model pages for these specific purpose pages:

- **Use a `key` field on the Page model**: this would be an optional unique field where editors can select from pre-defined keys (`home`, `search`, etc). Benefit is that it's a small change and can be a generic solution (record IDs are a system reference, slugs are url reference, keys are a developer reference). Disadvantage is that it's hard to ensure each special purpose page exists (editors could delete them). And if we offer multiple models there's no built-in way to keep keys unique across models.
- **Link special pages from the `app` model**: we could add a section for "Special pages" with required links for each specific purpose page. Benefits are that there's a clear place, and that we can enforce that each page is required and unique. The `app` model is also written to disk in a `prep` script, so it's easy for parts of the codebase to synchronously access it. For example for getting the search path needed in the `opensearch.xml` endpoint.

jurgenbelien:
# Changes

- Lets ESLint allow rest operator for removing properties from objects.
- Another update to error reporting
- Pass noIndex to Pages Content Collection entries
- Add purpose field to Pages Content Collection entries
- Migrates HomePage and NotFoundPage singletons to regular Pages
  -  These pages are set in Website (`app`) settings
  - Two separate migrations, one for schema, one for content make this easier
- Download said settings to `site.json`
- Creates a SpecialPages Contentent Collection for 404, home, and others pages to come
- Drops the Home route, now handled by `pages/[locale][...path]/index.astro`

# Associated issue

Resolves #291

# How to test

# Checklist


config/datocms/migrations/1752588127_featSpecialPages.ts
jurgenbelien, line 1: This is the schema migration

config/datocms/migrations/1752588128_featSpecialPagesContent.ts
jurgenbelien, line 1: This is the content migration.

MarleenEliza (commented):
Looks good!

But I have some concerns:
1. The slug of a page cannot be left empty. Isn't this desirable for the Home Page though? The language buttons in the top right corner still reflect the link with the set slug.
Screenshot 2025-07-21 at 11 28 58
Screenshot 2025-07-21 at 13 21 01

2. If the specialized pages are unset, the website cannot be built. I know this is by design, but this might be worth mentioning in the docs?

jurgenbelien:
> Looks good!
> 
> But I have some concerns:
> 
> 1. The slug of a page cannot be left empty. Isn't this desirable for the Home Page though? The language buttons in the top right corner still reflect the link with the set slug.
I resolved the issue with the language toggle, and added a test scenario for it. The slug field now has an help text explaining that for home it is ignored. 
> 2. If the specialized pages are unset, the website cannot be built. I know this is by design, but this might be worth mentioning in the docs?
I've added it to `getting-started.md`

config/datocms/migrations/1752588128_featSpecialPagesContent.ts
```diff
@@ -0,0 +1,196 @@
...
+    const { id, type, item_type, meta, creator, ...content } = record;
+    const title = isLocalizedField(content.title)
+      ? content.title[locales[0]]
+      : content.title;
+    console.log(`Updating record "${title || id}"`);
+    await client.items.update(
+      record.id,
+      replaceReferences(content, homePageModel.singleton_item.id, home_page),
+    );
+  }
```
jurgenbelien, line 169: this migrates any record that has a reference to the old home page to the new home page

config/datocms/migrations/1752588128_featSpecialPagesContent.ts
```diff
@@ -0,0 +1,196 @@
...
+    const { id, type, item_type, meta, creator, ...content } = record;
+    const title = isLocalizedField(content.title)
+      ? content.title[locales[0]]
+      : content.title;
+    console.log(`Updating record "${title || id}"`);
+    await client.items.update(
+      record.id,
+      replaceReferences(content, notFoundPageModel.singleton_item.id, not_found_page),
+    );
+  }
```
jurgenbelien, line 189: same for the not found pages, although the chance of that being necessary is slim

config/datocms/migrations/1752588127_featSpecialPages.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  console.log(
+    'Update Single link field "Not Found" (`not_found_page`) in model "\uD83D\uDDA5\uFE0F Website" (`app`)',
+  );
+  await client.fields.update(notFoundField.id, { position: 1 });
+  
+  console.log(
+    'Update Slug field "Slug" (`slug`) in model "\uD83D\uDCD1 Page" (`page`)',
+  );
+  await client.fields.update('SFfpl7xiQaGe5-jy6tlPfg', {
```
bazottie, line 109:
```suggestion
  await client.fields.update('page::slug', {

```

jbmoelker:
Stale and outdated. Will remain on the roadmap. Closing for now.

PR #297, opened 2025-07-24
Chore: Improve icon component

luukbrauckmann:
# Changes

- Refactor Icon component. I think this makes it a bit nicer.
- Add fill property to the icon CSS
- Remove `xlink:href`, because it never gets used by any browser and it is kinda "deprecated" feature.

My first idea was to implement the <title> element, but after some research I found out the title element isn't consistently supported by screenreaders and assistive technologies, even though it's part of the SVG specification. So I kept the aria-label approach but cleaned up the implementation.

# Associated issue

N.A.

# How to test

# Checklist

### DIFF
--- a/src/components/Icon/Icon.astro
+++ b/src/components/Icon/Icon.astro
@@ -3,9 +3,9 @@ import type { HTMLAttributes } from 'astro/types';
 import type { IconName } from '@assets/icon-sprite';
 export { iconNames } from '@assets/icon-sprite';
 
-type Props = {
+interface Props extends HTMLAttributes<'svg'> {
   name: IconName;
-} & HTMLAttributes<'svg'>;
+};
 
 const { 
   name,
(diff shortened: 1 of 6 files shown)


src/components/Icon/Icon.astro
```diff
@@ -7,19 +7,21 @@ interface Props extends HTMLAttributes<'svg'> {
...
-<svg { ...attributes } data-icon={name}>
-  <use href={`#${name}`} xlink:href={`#${name}`} />
+<svg 
+  data-icon={name} 
+  role={ariaLabel ? 'img' : 'presentation'} 
+  aria-hidden={!ariaLabel} 
+  aria-label={ariaLabel} 
+  {...props}
+>
+  <use href={`#${name}`} />
```
bazottie, line 24: Are these not identical to the previous implementation besides removing `xlink:href`?

PR #306, opened 2026-01-20
Edit page link for preview bar

anareyna:
# Changes
Adds an “edit page” link to the preview bar that links to the current page record editor (when we have id + __typename), so editors can jump from a preview page straight into the CMS record.

## Documentation link
See: https://github.com/voorhoede/head-start/blob/feat/edit-page/docs/preview-mode.md#edit-in-datocms-link 

# How to test

SCR-20260120-nsbz

# Checklist

### DIFF
--- a/src/components/PreviewMode/PreviewMode.client.ts
+++ b/src/components/PreviewMode/PreviewMode.client.ts
@@ -113,17 +113,18 @@ class PreviewMode extends HTMLElement {
   }
 
   connectedCallback() {
-    const isLinkableRecord = this.editableRecord?.id && this.editableRecord?.type && this.#datocmsProject;
-    if (!isLinkableRecord) {
+    const record = this.editableRecord;
+    const isLinkableRecord = record?.id && record?.type && this.#datocmsProject;
+    if (!isLinkableRecord || !record) {
       return;
     }
 
-    const itemTypeId = PreviewMode.getItemTypeId(this.editableRecord?.type);
+    const itemTypeId = PreviewMode.getItemTypeId(record.type);
     if (!itemTypeId) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/datocms/modelApiKeys.ts
jurgenbelien: Should this be committed if it is generated from the associated DatoCMS environment? Consider adding it to `.gitignore`

src/lib/generated-files.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '@lib/datocms/itemTypes.json' {
+  const value: Record<string, { id: string }>;
+  export default value;
+}
+
```
jurgenbelien: I'm expecting Typescript to treat the contents of the JSON as if it were a variable "`as const`". Take a look at `@src/lib/i18n/index.ts`. For me, Typescript correctly infers that the messages import there results in an object with key `en` and `nl`.

src/components/PreviewMode/features/EditInDatoCms/EditInDatoCms.astro
```diff
@@ -0,0 +1,26 @@
...
+      edit in DatoCMS
+    </a>
+  )
+}
+
+<style>
+  a {
+    color: currentColor;
+  }
+</style>
```
jurgenbelien: Sure this is necessary? I would expect this styling to be the case out of the box

jurgenbelien (commented):
If it weren't for how the PreviewMode works now I think this is a great way to do it. However, since there is already a lot of plumbing in the PreviewMode I'd rather leverage that, meaning showing links to records that are being watched

jurgenbelien (commented):
Looks good but the type reassignments could hint at an underlying issue. Could you explain why this is necessary? I'd rather have no type reassignments at all honestly.

jurgenbelien (approved):

PR #315, opened 2026-02-12
feature: Animated Counter Block

### LINKED ISSUE #311: Inline Counter Block
## User story

**As a marketeer,
I want to add animated counters to the website,
to impress visitors with our numbers.**

## Requirements

- Component should accept a numeric value.
- Component should be injectable in a text field (Inline Block).
- Value is animated from 0 to its actual value in a given time (configurable with default?).
- Value is automatically formatted in the active locale with correct decimal and grouping separators.
- Numbers within the value don't shift left/right while animating (use `tabular-nums`).
- Content around the component doesn't shift while animating (reserve space of eventual value).
- Value only animates when it's in view.
- Value only animates when user is okay with animations.

## Notes

Component could also be (re)used to compose other types of content blocks with more specific features like a prefix, suffix, title, text, etc. These features don't need to be in the core counter component.

## References

- [Counter PoC](https://github.com/voorhoede/head-start/compare/main...feat/counter-component) by @jbmoelker 
- [Fudura](https://www.fudura.nl/nl/) tip by @milailijevic:

Image

milailijevic:
> [!IMPORTANT] 
~Depends on #316~ #316 is merged
# Changes
- Adds an animated Counter Block 

# Associated issue

Resolves #311 

# How to test
https://feat-counter-block.head-start.pages.dev/en/demos/counter-block-demo/

# Checklist

### DIFF
--- a/config/datocms/migrations/1773320919_featCounterBlock.ts
+++ b/config/datocms/migrations/1773320919_featCounterBlock.ts
@@ -40,4 +40,35 @@ export default async function (client: Client) {
     },
     default_value: null,
   });
+
+  console.log('Update existing fields/fieldsets');
+
+  console.log(
+    'Update Structured text field "Text" (`text`) in block model "\uD83D\uDCDD Text Block" (`text_block`)',
+  );
+  await client.fields.update('NtVXfZ6gTL2sKNffNeUf5Q', {
+    validators: {
+      required: {},
+      structured_text_blocks: {
+        item_types: [
+          'F60ZY1wFSW2qbvh99poj3g',
+          'QYfZyBzIRWKxA1MinIR0aQ',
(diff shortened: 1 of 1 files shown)


jbmoelker:
Nice setup. 

I would right align (`text-align: end`) values so all digits stay in the same place.
However I see that even if I do that the text after the counter still jumps after page load. So I think that needs to be fixed. See video below.

And good to test this with numbers with decimals as well! I see that the animating value doesn't have decimals. I think that value should include the same number of decimals as the eventual number. So that also numbers with decimals have all digits stay in the same place.

https://github.com/user-attachments/assets/8d1ceb0b-982a-4fc8-b13d-8a81b583ec61

PR #335, opened 2026-04-01
Use datocms astro structured text instead of own structured text component

### LINKED ISSUE #333: Replace own components with those from @datocms/astro
- StructuredText -> [docs/astro/structured-text-fields](https://www.datocms.com/docs/astro/structured-text-fields)
- Image ->  [voorhoede/Image](https://github.com/voorhoede/Image)
- QueryListener ->  [docs/astro/real-time-updates](https://www.datocms.com/docs/astro/real-time-updates)

JopMolenaar:
# Changes

Use datocms astro components instead of own components
- datocms/astro for structuredText [structuredText docs/astro](https://www.datocms.com/docs/astro/structured-text-fields)
- Updates docs 

# Associated issue

Resolves #333

# How to test

1. Check if nothing breaks when making content in a textblock in random blocks where this component is used. 
2. Compare it to the main branch (changes in primary and view it in preview)

# Checklist


src/blocks/ImageBlock/Image.astro
```diff
@@ -1,99 +1,83 @@
...
+
+const astroImageData = {
+  src: image.url,
+  srcSet: image.responsiveImage?.srcSet ?? null,
+  title: image.title,
+  width: image.width,
+  height: image.height,
+  alt: altText,
+  base64: image.responsiveImage?.base64 || '',
+};
```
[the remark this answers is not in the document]
JopMolenaar: @milailijevic  Yes, `TextImageBlock.fragment.graphql` is now also fetching srcSet, and if I search on `responsiveImage`, all of them have srcSet in it. 
But even when srcSet is missing, AstroImage has its own `buildSrcSet` function, so it won't break anything.

src/blocks/ImageBlock/ImageBlock.fragment.graphql
milailijevic, line 1: maybe using imgixParams here to limit the full size of it and set the q to 80 as we previously had
```
fragment ImageBlock on ImageBlockRecord {
  id
  image {
      responsiveImage( imgixParams: { auto: [format, compress], q: 80, fit: max, w: 1200} ) {
      srcSet
      src
      base64
      alt
      title
      width
      height
    }
  }
}

```

and then just use `const astroImageData = image.responsiveImage ?? null; `in Image.astro.

This is just an idea, it scales the image with **dpr** as in this example but limiting the biggest one is the issue here.
https://www.datocms-assets.com/145765/1730394116-head-start-cover.jpg?auto=format%2Ccompress&dpr=0.25&fit=max&q=80&w=1200

The [maxW param ](https://docs.imgix.com/en-US/apis/rendering/size/maximum-width) doesn't work without the fit=crop

having imgixParams: { auto: [format, compress], q: 80, fit: max, w: 1200} in the query works fine, but the params stay the same except from the **dpr**

src/blocks/TextImageBlock/TextImageBlock.fragment.graphql
milailijevic, line 62: should follow the same structure as in ImageBlock.fragment.graphql if you change it

src/components/LinkToRecord/LinkToRecord.astro
```diff
@@ -1,24 +1,32 @@
...
 import { getLocale } from '~/lib/i18n';
 import { getHref, type RecordRoute } from '~/lib/routing';
 import Link from '~/components/Link/Link.astro';
 import LinkToFile from '~/components/LinkToFile/LinkToFile.astro';
 
-export type Props = Omit<HTMLAttributes<'a'>,'href'> & {
+export type Props = Omit<HTMLAttributes<'a'>, 'href'> & {
   openInNewTab?: boolean;
-  record: RecordRoute;
+  record: RecordRoute | VariableRecord;
```
milailijevic, line 11: Rebase to main, the VariableRecord type is not used here anymore

milailijevic (changes requested):

src/blocks/TextImageBlock/TextImageBlock.fragment.graphql
```diff
@@ -68,6 +68,7 @@ fragment TextImageBlock on TextImageBlockRecord {
     responsiveImage {
       aspectRatio
       base64
+      srcSet
```
milailijevic: If we won't handle the datocms image in this PR is this necessary in the query?

milailijevic (dismissed):

anareyna (approved):

PR #351, opened 2026-04-17
Replace own custom fonts loader with Astro Fonts Api

### LINKED ISSUE #350: Replace own custom fonts loader with Astro Fonts API
Replace our own font-loader with the [Astro Fonts API](https://docs.astro.build/en/reference/font-provider-reference/)
Its officially implemented in astro v6.0, but Jasper says its fine to implement it now under experimental in de the config.

JopMolenaar:
# Changes

- Implements the new Astro Fonts API
- Removes our own fonts loader

# Associated issue

Resolves #350 

# How to test

1. Open the preview link
2. Open inspector and check the fonts that are used.
3. Maybe even experiment using other fonts of fontsource by adding them to the config and use them. 
4. Check the if the editor guide also has the right font. 

# Checklist


decrek (changes requested):
This PR changes how fonts are loaded for the end user. Previously fonts were preloaded, but in the preview link the font face declarations are just inlined. I would say that this is worse than before regarding loading experience of fonts. At least it is different. If you want to just change how fonts are loaded, probably adding `preload` on the font is enough? See https://docs.astro.build/en/guides/fonts/#preloading-fonts. Im all for using this build in method btw, I just think the output should be the same

src/layouts/PerfHead/PerfHead.test.ts
```diff
@@ -16,25 +15,28 @@ describe('PerfHead', async () => {
     });
   });
 
-  test('preloads font files for faster font loading and page rendering', () => {
```
decrek, line 19: I would keep this test!

src/layouts/PerfHead/PerfHead.astro
```diff
@@ -23,15 +21,4 @@ const fontFaceDeclaration = fonts.map((font) => getFontFaceDeclaration(font)).jo
...
-      rel="preload"
-      as="font"
-      type="font/woff2"
-      href={url}
-      crossorigin="anonymous"
-    />
-  ))
-}
-<style set:html={fontFaceDeclaration}></style>
+<Font cssVariable="--font-archivo" />
```
decrek: try adding preload attribute here

JopMolenaar:
@decrek `preload ` seems to be the fix and is now doing the same as on [https://head-start.pages.dev/en/](https://head-start.pages.dev/en/).

anareyna (approved):

PR #345, opened 2026-04-08
feat: expose content pages in markdown format

slimluccii:
## Changes

- Adds a runtime API endpoint (`/api/content/[...path].md`) that serves content pages as Markdown, enabling AI agents and LLMs to consume site content in a clean, structured format
- Converts HTML to Markdown using `unified` + `rehype-remark` + `remark-stringify`, scoped to the `<main>` content area only
- Gates the endpoint on the existing CMS "allow AI bots" setting — returns 404 when AI bots are not allowed
- Adds an in-memory cache (5 min TTL) for markdown responses to avoid repeated HTML-to-Markdown conversion
- Propagates `noindex` from page-level robots meta to an `X-Robots-Tag: noindex` response header
- Adds `<link rel="alternate" type="text/markdown">` to the HTML `<head>` so crawlers and agents can discover the Markdown version
- Adds `hast-util-select`, `rehype-remark`, `remark-gfm`, `remark-stringify`, and `unified` as dependencies

## Associated issue

Resolves #340 

## How to test

1. Enable "Allow AI bots" in the CMS app settings
2. Open any content page (e.g. `/en/some-page/`)
3. Inspect the HTML `<head>` — verify a `<link rel="alternate" type="text/markdown" href="/api/content/en/some-page.md">` tag is present
4. Visit `/api/content/en/some-page.md` directly — verify it returns the page content as clean Markdown with `Content-Type: text/markdown`
5. Verify the Markdown only contains the main content area (no nav, footer, etc.)
6. Disable "Allow AI bots" in the CMS — verify the `.md` endpoint returns 404 and the `<link>` tag is no longer rendered
7. Set a page to `noindex` — verify the `.md` endpoint includes the `X-Robots-Tag: noindex` header

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have made sure that my PR is easy to review (not too big, includes comments)
- [ ] ~I have made updated relevant documentation files (in project README, docs/, etc)~
- [ ] ~I have added a decision log entry if the change affects the architecture or changes a significant technology~
- [x] I have notified a reviewer

PR #354, opened 2026-04-22
Open page in llm copy button with dropdown

JopMolenaar:
# Changes

- Adds copy page button on every page if AllowAIBots is true
  - Which copies the the markdown version of the current page. #345
- Adds llmLinks in a popover next to the copy button, if these are configured in the CMS
  - These llm links will open a llm with the prompt you've added in the CMS. 
  - Otherwise it will only add the pageUrl as promp
- Adds migration file for added cms options
- Adds translations // TODO

# Associated issue

Resolves #341

# How to test

1. Open a page
2. Copy the page and paste it somewhere to verify that it has copied the right content. 
3. Test errors. What if the fetch fails, remove the alternate link for example. 
4. Change CMS settings like AllowAIBots in order to check if the button is not rendered in this case.  
3. Check the popover on accessibility 
4. Open the page in llm via the popover link

# Checklist


src/components/OpenInLlm/OpenInLlm.astro
```diff
@@ -0,0 +1,90 @@
+---
+import Icon from '~/components/Icon/Icon.astro';
+import { t } from '~/lib/i18n';
+
+export type LlmOption = {
```
slimluccii: This could be the ExternalLinkFragment type

src/components/OpenInLlm/OpenInLlm.client.ts
```diff
@@ -0,0 +1,91 @@
...
+      const text = await this.#buildCopyText(url);
+      await navigator.clipboard.writeText(text);
+      this.#setLabel(this.#copyButton?.dataset.labelCopied ?? '');
+    } catch {
+      this.#setLabel(this.#copyButton?.dataset.labelError ?? '');
+    }
+    setTimeout(() => this.#setLabel(this.#originalLabel), 2000);
+  };
+
+  #buildCopyText = async (pageUrl: string): Promise<string> => {
```
slimluccii, line 96: When I click Copy page. The button says 'Failed to copy page'. This error however is a bit obscure and I think it would be nice to add a console error for this with more info about what went wrong.

src/components/OpenInLlm/OpenInLlm.astro
```diff
@@ -0,0 +1,84 @@
...
+  llmOptions: ExternalLinkFragment[];
+  prompt: string;
+};
+
+const { llmOptions, prompt } = Astro.props;
+
+const popoverId = 'open-in-llm-popover';
+const pageUrl = Astro.url.href;
+
+let searchText = prompt.replace('{ pageUrl }', pageUrl);
```
slimluccii: `.replace('{ pageUrl }')` is really fragile. I think you should implement a more robust way. Maybe a regex key? Or third party package? 

Also this replaces the first instance of the `{ pageUrl }`. What if someone wants to add multiple `{ pageUrl }{ pageUrl }{ pageUrl }`? You could handle that through a regex or `.replaceAll('{ pageUrl }')` for example.

src/components/OpenInLlm/OpenInLlm.client.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!popover || !this.#buttonGroup) return;
+
+    const placement = (popover.getAttribute('data-placement') ??
+      'bottom-start') as Placement;
+    computePosition(this.#buttonGroup, popover, { placement }).then(
+      ({ x, y }) => {
+        Object.assign(popover.style, {
+          left: `${x}px`,
+          top: `${y}px`,
+          width: `${this.#buttonGroup!.offsetWidth}px`,
```
slimluccii: I think you should remove the width limitation since this causes every line to break because the width of the button set is always way smaller probably. If you're trying to prevent the box to flow off screen then I think you should alter your positioning to detect that and adjust the position.

slimluccii (changes requested):

slimluccii (approved):

PR #357, opened 2026-04-29
feat: llms.txt

slimluccii:
# Changes

- Add `/llms.txt` endpoint exposing site name, summary, intro, and full menu tree for LLM crawlers (follows [llmstxt.org](https://llmstxt.org/) spec)
- Add `llms_intro` field to the Website model in DatoCMS (markdown, AI section fieldset) with `{{ siteName }}` interpolation support
- Hide page list when `allowAiBots` is disabled; default `llms_intro` aligns with that state
- Add `llmsTxt` builder + `renderLlmsTxtItems` helper in `src/lib/seo` with unit tests
- Document the `/llms.txt` endpoint in `docs/seo.md`
- Bump DatoCMS environment to `feat-develop-sandbox-15-04`

# Associated issue

#343 

# How to test

1. Open preview link
2. Visit `/llms.txt`
3. Verify the response contains:
   - `# {siteName}` heading
   - `> {siteSummary}` blockquote (when set)
   - Intro paragraph with `{{ siteName }}` replaced by the actual site name
   - `## Pages` section listing the full menu tree (only when AI bots are allowed)
4. In DatoCMS, toggle `allowAiBots` off → reload `/llms.txt` → verify `## Pages` section is gone and intro reflects disabled state
5. Edit `llms_intro` in the CMS → verify the change appears in the output

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have made sure that my PR is easy to review (not too big, includes comments)
- [x] I have made updated relevant documentation files (in project README, docs/, etc)
- [ ] ~~I have added a decision log entry if the change affects the architecture or changes a significant technology~~
- [x] I have notified a reviewer

PR #377, opened 2026-05-15
fix: post-merge feedback batch (OpenInLlm placement, .md links, copy duplicate)

slimluccii:
# Changes

- Move `OpenInLlm` out of `<main>` into a `.page-context` container next to the breadcrumbs, wrapped with `datocmsNoIndex` + `data-nosnippet` + `<!--googleoff: index-->` / `<!--googleon: index-->` comments so it is not indexed and no longer leaks into the generated markdown alternative of pages.
- Extract reusable `htmlToMarkdownPath` / `markdownToHtmlPath` / `isMarkdownApiPath` helpers in `src/lib/routing/markdown.ts` (with round-trip tests). Refactor `rehypeRewriteLinksToMarkdown` to use them. External, `mailto:`/`tel:`, hash-only and `/api/*` links are left alone.
- Rewrite internal page URLs in `/llms.txt` to point to the `.md` alternative so consumers stay in the markdown world.
- Drop the page URL that was prepended before the markdown body in the "Copy page" output — the URL is already in the frontmatter.
- Add an **Open as Markdown** entry as the first item in the LLM dropdown (linking to the `.md` alternative of the current page). New `open_as_markdown` translation key (en/nl).
- Use the locale-aware `openPageInLlmPrompt` / `openPageInLlmOption` values via the `DefaultLayout` GraphQL query (instead of the global `app` singleton at the default locale), so the prompt and LLM list are translated per locale.

# Associated issue

Follow-up on feedback for the merged llms.txt + Open in LLM work (#357, #374). No single issue.

# How to test

1. Open the preview deploy for this branch.
2. Visit any content page (e.g. `/en/`).
3. Confirm the "Copy page" + LLM dropdown render right of the breadcrumbs, not inside the article body.
4. Open the LLM dropdown — the first entry is **Open as Markdown** and links to `/api/content/<path>.md`.
5. Fetch the markdown alternative directly (e.g. `/api/content/en.md`) and verify:
   - Internal links are rewritten to `<origin>/api/content/<path>.md`.
   - External links (astro.build, datocms.com, …) are untouched.
   - No "Copy page" / "Open in Claude" / "Open in ChatGPT" entries leak into the body.
   - Frontmatter contains `url` once; the markdown body does not repeat it.
6. Click "Copy page" on the rendered page and paste in a text editor — output should match `/api/content/<path>.md` exactly.
7. Fetch `/llms.txt` and verify the internal page links point to `/api/content/<path>.md` (external menu items still use their absolute URLs).
8. Switch to a non-default locale (e.g. `/nl/`) and confirm the LLM dropdown's prompt and option titles are localized.
9. View source on a page where AI bots are allowed and confirm the `OpenInLlm` block is wrapped in `<!--googleoff: index-->` / `<!--googleon: index-->` and has `data-nosnippet`.

# Checklist

- [ ] ~~I have made updated relevant documentation files (in project README, docs/, etc)~~ — no doc surface affected
- [ ] ~~I have added a decision log entry if the change affects the architecture or changes a significant technology~~ — no architectural change


src/components/OpenInLlm/OpenInLlm.astro
```diff
@@ -21,13 +23,15 @@ const llmLinks = llmOptions.map((option) => {
   href.searchParams.set('q', searchText);
   return { ...option, href: href.toString() };
 });
+
+const hasDropdown = true;
```
JopMolenaar: If this is always true we can remove the checks right?

src/components/OpenInLlm/OpenInLlm.astro
```diff
@@ -12,6 +13,7 @@ const { llmOptions, prompt } = Astro.props;
 
 const popoverId = 'open-in-llm-popover';
 const pageUrl = Astro.url.href;
+const markdownHref = `${Astro.url.origin}${htmlToMarkdownPath(Astro.url.pathname)}`;
 
 let searchText = prompt.replaceAll(/\{\s*pageUrl\s*\}/g, pageUrl);
 if (searchText === '') searchText = pageUrl;
```
JopMolenaar: Can this pageUrl become the markdown page url so that the LLM opens the md version directly?

src/components/OpenInLlm/OpenInLlm.client.ts
JopMolenaar, line 5: Can these `| null = null`'s be deleted by adding a constructor instead?
```
constructor() {
    super();
    this.#copyButton = this.querySelector('[data-copy-button]') as HTMLButtonElement;
```

slimluccii: I think the `connectedCallback` is the right place to define the element, because it gets called when the component is fully mounted.

JopMolenaar:
@slimluccii In the test steps of this pr it says: 
"Internal links are rewritten to <origin>/api/content/<path>.md."

But when I check the markdown version of the homepage I see links like
`* [Accessibility (a11y)](https://38f2d5e1.head-start.pages.dev/api/content/en/documentation/accessibility.md)`

While these links share the same origin. 
Is this intended?

slimluccii:
> @slimluccii In the test steps of this pr it says: "Internal links are rewritten to /api/content/.md."
> 
> But when I check the markdown version of the homepage I see links like `* [Accessibility (a11y)](https://38f2d5e1.head-start.pages.dev/api/content/en/documentation/accessibility.md)`
> 
> While these links share the same origin. Is this intended?

Yes using absolute url's for agents is a safer bet and prevents the agent from resolving to a different base.

JopMolenaar (approved):

PR #404, opened 2026-05-22
Remove unused globby dependency

slimluccii:
globby was added in Oct 2023 for scripts/create-404-pages.mjs, which no longer exists. Nothing in the codebase imports it. Removing it also reclassifies its transitive deps (fast-glob, braces, etc.) as dev-only.


JopMolenaar:
Approved!

PR #418, opened 2026-05-26
feat: ai search component

anareyna:
Depends on https://github.com/voorhoede/head-start/pull/379

Adds the "Ask AI" page so people can actually use the AI search we wired up in the previous PR. Type a question, get an answer with sources.

Screenshot 2026-05-27 at 09 21 26

## What's added

- A new **Ask AI** page at `/[locale]/ask/`. Question box at the top, answer streams in below as it's written, sources show up alongside it.
- An **Ask AI** link in the main nav, right next to Search. Same link in the mobile nav.
- The answer renders as proper markdown while it's still streaming, so code blocks and lists look right from the first chunk, not just when it finishes.
- You can share a question via URL: `/en/ask/?query=...` runs the search on page load, and submitting the form keeps the URL in sync. Good for linking people to "here's the answer I got".

## How to test

Check cloudfare preview link https://feat-ai-search-component.head-start.pages.dev/en/ask  or

1. Run the app locally (env vars from the [previous PR](ttps://github.com/voorhoede/head-start/pull/379) need to be set).
2. Visit `/en/ask/`. Ask something the site would know about. You should see "Thinking…", then the answer appear word by word, with the sources it used.
3. Try a deep link like `/en/ask/?query=How%20do%20I%20add%20a%20new%20locale%3F`. It should run on its own.
4. Check the item is in header nav: "Ask AI" 

## Checklist

- [x] Self-reviewed
- [x] Easy to review (one component, one page, one nav link)
- [ ] Notified a reviewer

### DIFF
--- a/src/components/AiSearch/AiSearch.astro
+++ b/src/components/AiSearch/AiSearch.astro
@@ -1,4 +1,6 @@
 ---
+import { t } from '~/lib/i18n';
+import { siteName } from '~/lib/seo';
 import { queryParamName } from '~/lib/search';
 
 /*
@@ -9,39 +11,39 @@ const inputId = 'ai-search-input';
 ---
 
 <ai-search
-  data-thinking-text="Thinking…"
-  data-empty-text="No answer found for that question."
-  data-error-text="Something went wrong. Please try again."
+  data-thinking-text={t('ask_thinking')}
+  data-empty-text={t('ask_empty')}
+  data-error-text={t('ask_error')}
 >
   <div class="ai-search__head">
-    <h1>Ask Head Start</h1>
+    <h1>{t('ask_on_site', { siteName: siteName() })}</h1>
     <p class="ai-search__intro">
-      Ask a question and get an answer drawn from this site's content.
+      {t('ask_intro')}
     </p>
     <form class="ai-search__form" role="search">
-      <label class="a11y-sr-only" for={inputId}>Ask a question</label>
+      <label class="a11y-sr-only" for={inputId}>{t('ask_label')}</label>
       <input
         class="ai-search__input"
         id={inputId}
         name={queryParamName}
         type="search"
-        placeholder="Ask a question about this site…"
+        placeholder={t('ask_placeholder')}
         autocomplete="off"
         required
         data-input
       />
-      <button type="submit" data-submit>Ask</button>
+      <button type="submit" data-submit>{t('ask_submit')}</button>
     </form>
   </div>
 
   <div class="ai-search__result" data-result hidden>
     <article>
-      <h2>Answer</h2>
+      <h2>{t('ask_answer')}</h2>
       <p class="ai-search__status" data-status aria-live="polite"></p>
       <div class="ai-search__text" data-answer></div>
     </article>
     <aside data-sources-wrap hidden>
-      <h2>Sources</h2>
+      <h2>{t('ask_sources')}</h2>
       <ul class="ai-search__sources" data-sources></ul>
     </aside>
   </div>
--- a/src/components/AiSearch/AiSearch.astro
+++ b/src/components/AiSearch/AiSearch.astro
@@ -80,11 +80,6 @@ const inputId = 'ai-search-input';
     background: #f5f5f5;
     font-size: 0.8em;
   }
-  ai-search .ai-search__sources {
-    list-style: none;
-    margin: 0;
-    padding: 0;
-  }
   ai-search .ai-search__sources li {
     margin-block-end: 0.75rem;
   }
--- a/src/components/AiSearch/AiSearch.astro
+++ b/src/components/AiSearch/AiSearch.astro
@@ -1,0 +1,96 @@
+---
+import { queryParamName } from '~/lib/search';
+
+/*
+ * <ai-search> component: the question box, the streamed answer, and its sources.
+ */
+
+const inputId = 'ai-search-input';
+---
+
+<ai-search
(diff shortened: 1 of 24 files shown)


src/components/AiSearch/AiSearch.astro
```diff
@@ -0,0 +1,91 @@
...
+
+<ai-search
+  data-thinking-text="Thinking…"
+  data-empty-text="No answer found for that question."
+  data-error-text="Something went wrong. Please try again."
+>
+  <div class="ai-search__head">
+    <h1>Ask Head Start</h1>
+    <p class="ai-search__intro">
+      Ask a question and get an answer drawn from this site's content.
```
JopMolenaar: Could you add the english text as translations? Or maybe some translations and other text that content creators would change a lot, such as "Ask a question and get an answer drawn from this site's content." can be placed in the dato website model -> AI section?

Image

src/components/AskLink/AskLink.astro
```diff
@@ -0,0 +1,7 @@
+---
+import { getLocale } from '~/lib/i18n';
+
+const askUrl = `/${getLocale()}/ask/`;
```
JopMolenaar: Should this page show on:
- en/ask
- nl/vraag

And the slug should be configurable per added locale?

src/pages/[locale]/ask/index.astro
```diff
@@ -0,0 +1,26 @@
...
+const pageUrls = locales.map((locale) => ({ locale, pathname: getAskPathname(locale) }));
+const breadcrumbs: Breadcrumb[] = [{ text: pageTitle, href: getAskPathname(locale) }];
+---
+
+<Layout
+  breadcrumbs={breadcrumbs}
+  pageUrls={pageUrls}
+  seoMetaTags={[noIndexTag, titleTag(pageTitle)]}
+>
+  <AiSearch />
```
Comment (JopMolenaar) [inline src/pages/[locale]/ask/index.astro:31]: Should this page still be visible if Allow AI Bots is turned off in dato? When Allow AI Bots is turned off, markdown pages will not be generated and all the other AI features will also not appear.

JopMolenaar (changes requested):

src/components/AskLink/AskLink.astro
anareyna: good question, for now I went with the same approach as search: a fixed /ask/ slug for every locale, so `/en/ask/` and `/nl/ask/`. I pulled it into a getAskPathname() helper, same idea as getSearchPathname().

CMS pages get localized slugs, but utility pages like search don't, and ask fits that for now

src/pages/[locale]/ask/index.astro
Comment (anareyna) [reply src/pages/[locale]/ask/index.astro:31]: good catch, no it shouldn't be visible when Allow AI Bots is off. I've added the nav link and the page itself behind the same allowAiBots check we use for Open in LLM and the markdown endpoint

src/components/AiSearch/AiSearch.astro
anareyna: i went with translations, same as the search page, with en + nl in a migration, I didn't put the intro in the Website AI section for now. We can add a cms field later if editors need to change that per project.

src/components/AiSearch/AiSearch.client.ts
```diff
@@ -0,0 +1,171 @@
...
+    } catch (error) {
+      console.error('ai-search:', error);
+      this.#status.textContent = this.dataset.errorText ?? '';
+    } finally {
+      this.#setBusy(false);
+    }
+  }
+
+  #addSource(title: string, url: string) {
+    if (!this.#sources || !this.#sourcesWrap) return;
```
JopMolenaar: `#sourcesWrap` is already checked in the guard of `#run` so one of them could be removed

src/components/AiSearch/AiSearch.client.ts
```diff
@@ -0,0 +1,171 @@
...
+
+class AiSearch extends HTMLElement {
+  #form: HTMLFormElement | null = null;
+  #input: HTMLInputElement | null = null;
+  #submit: HTMLButtonElement | null = null;
+  #result: HTMLElement | null = null;
+  #status: HTMLElement | null = null;
+  #answer: HTMLElement | null = null;
+  #sourcesWrap: HTMLElement | null = null;
+  #sources: HTMLElement | null = null;
```
JopMolenaar: These `| null = null`'s  confuses me because these elements are hardcoded and they all exist?

I don't know the right approach and it is implemented in different ways in different files. 
But maybe removing the null's and using a constructor is cleaner?

``` 
constructor() {
    super();
    this.#form = this.querySelector('form') as HTMLFormElement;
```

anareyna: yea makes sense 👍🏽  matched the constructor pattern from AppMenu.client.ts, the nulls were noise

JopMolenaar:
@anareyna Nice, looks good. Could you update the datocmsEnvironment to the primary env? I want to test something but the /ask page is giving a 404.

anareyna:
> #379

@JopMolenaar thanks, i updated this and the parent branch https://github.com/voorhoede/head-start/pull/379 with the latest, just fyi when testing the ask page wont be visible if allowAiBots is off or if the translations are missing, so perhaps you can test and run migrations in a temp new dato env?

JopMolenaar (approved):
I would not merge it yet in feast-ai-search because that one is not yet approved/reviewed and it would make it more difficult. 
But it looks good! Nice work!

PR #424, opened 2026-06-01
feat: AI Chat Component 

anareyna:
## Depends on https://github.com/voorhoede/head-start/pull/418

# Changes

Screenshot 2026-06-01 at 09 04 35

Adds multi-turn **Chat** on top of the same Cloudflare AI Search backend the Ask AI page uses. Ask a question, then ask a follow-up that remembers the previous turn instead of starting over.

# Changes

- A new **Chat** page at `/[locale]/chat/`. Multi-turn against the same Cloudflare backend as Ask AI: streams the answer with sources, no new env vars.
- Separate `/api/ai-chat` endpoint instead of overloading search. Both stay live for now; which becomes the main surface is a later product call. The server caps history to the last 10 messages so token cost stays bounded.
- **Chat** link in the main and mobile nav, gated like Ask AI: only when AI bots are allowed and the translation exists. 
- Chat labels translated (en + nl) via a DatoCMS migration

# Associated issue

n/a

# How to test

## Setup (skip if you already have it running)

The AI Search backend has to be configured or the endpoint returns a 503. Follow the **Enable AI Search** section in `docs/getting-started.md` (same vars the parent PR needs), and make sure AI bots are allowed for the site, otherwise the page 404s and the nav link hides on purpose.

If there's a Cloudflare preview deploy for this branch you can use that. The backend is already wired up there.

## Verify the change

1. Open `/en/chat/`. Ask something the site would know about. You should see "Thinking…", then the answer stream in word by word, with the sources it used underneath.
2. Ask a follow-up that only makes sense in context. It should answer relative to the earlier turn, not treat it as a fresh question.
3. Refresh the page. Your conversation should still be there.
4. Hit **Clear chat**. Everything disappears and stays gone after a refresh.
5. Switch to `/nl/chat/`. Your `en` conversation should not show there.
6. Check the nav: **Chat** is in the header next to Search, and in the mobile menu.
7. Turn off "Allow AI Bots" and reload. `/en/chat/` should 404 and the **Chat** nav link should be gone.

# Checklist

- [x] I have updated relevant documentation files (`docs/getting-started.md`)
- [x] I have added a decision log entry (`docs/decision-log/2026-05-27-ai-chat-multiturn.md`)


src/components/AiChat/AiChat.client.ts
```diff
@@ -0,0 +1,251 @@
...
+      return [];
+    }
+  }
+
+  #saveHistory() {
+    try {
+      localStorage.setItem(this.#storageKey(), JSON.stringify(this.#messages));
+    } catch {
+      // Private mode or quota full. Chat keeps working in memory; refresh
+      // would lose it, but that's better than blocking the user.
```
JopMolenaar: Is it an idea to add a text somewhere that says: 
"Chat not saved. Refreshing the page will lose earlier messages.." 
or
"This chat could not be saved. You can continue chatting, but if you refresh the page, earlier messages will be lost."

src/pages/api/ai-chat.ts
```diff
@@ -0,0 +1,115 @@
...
+
+// Cloudflare expects the first message to be from the user. If we trimmed off a
+// user turn and the slice now starts with assistant, drop messages until it
+// does. Otherwise the upstream call errors.
+const capHistory = (messages: ChatMessage[]) => {
+  let trimmed = messages.slice(-HISTORY_CAP);
+  while (trimmed.length > 0 && trimmed[0].role !== 'user') {
+    trimmed = trimmed.slice(1);
+  }
+  return trimmed;
```
JopMolenaar, line 55: Should we let the user know that when the chat goes beyond the 10 messages, old messages wont be taken in account with new messages?

JopMolenaar (commented):
Looks good! Would it also be cool (probably another pr) to keep the markdown version of this chat page up to date? So when you click the open page in a LLM, this LLM can read the chat

src/components/AiChat/AiChat.client.ts
anareyna: good call, added this:
Screenshot 2026-06-03 at 08 10 38
tested with `Storage.prototype.setItem = function () { throw new Error('forced quota'); };`

src/pages/api/ai-chat.ts
anareyna: the 10 message cap is a backend context window limit, the ui still shows the full transcript so nothing visibly disappears, warning that "earlier messages aren't considered" might confuse more than help

anareyna:
> Looks good! Would it also be cool (probably another pr) to keep the markdown version of this chat page up to date? So when you click the open page in a LLM, this LLM can read the chat

good idea but what I see is that our ` .md ` pattern mirrors server rendered cms content, but the chat is client-side only (it lives in localStorage, the server never sees the transcript). A client-side "export chat as markdown" could be a nice follow-up though :)

src/pages/api/ai-chat.ts
JopMolenaar: Yes true. But it doesn't have to be somewhere when the user directly sees it. Maybe a subtile icon with a little text directly on the messages that are out of the context window could work. Or a line like this:

- User chat
> ai chat
- user chat
> ai chat

---Messages above this line are out of the current context window---

- User chat
> ai chat
- user chat
> ai chat
- User chat
> ai chat
- user chat
> ai chat
- User chat
> ai chat
- user chat
> ai chat


😄 idk, just an idea

